PR #24, opened 2019-10-10
Reproducible nix build

jb55:
This adds dockerless reproducible nix builds.

Nix is a platform-agnostic way to do reproducible builds without needing a virtual machine.

I've tucked this away in a scripts subfolder, for devs who prefer non-docker builds.

This still needs a few things:

- [ ] Runtime dependencies


jb55:
these deps are kind of gnarly to make reproducible so I'm going to tackle this later

PR #26, opened 2019-10-11
Adding initial react generator

kb-commit:


kb-commit:
### Overall:
1. Added initial skeleton from create react app that we can evolve from
2. Instead of writing generator for file by file, added generator for directory, this should allow us to skip a lot of overhead in code
3. instead of using .tmpl using real extensions needed for 2 and still work like templates

### Planned next:
1. Add unit tests for all the above
2. Replace existing file by file generators to directory generators

templates/react/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "{{ .Name }}",
```
kb-commit: templating still works with real extensions

templator/templator.go
```diff
@@ -108,6 +114,48 @@ func NewDockerFileTemplator(box *packr.Box) *DockerTemplator {
 	return &DockerTemplator{
 		ApplicationDocker: appTemplate,
 		HttpGatewayDocker: httpTemplate,
-		DockerIgnore: ignoreTemplate,
+		DockerIgnore:      ignoreTemplate,
+	}
+}
+
+type DirectoryTemplator struct {
```
kb-commit: this provides a very generic templator that we can use for any directory

Pritesh-Patel: How would this work for bits where each file needs a bit of additional context like here https://github.com/commitdev/commit0/blob/f679a5cb7eedf23c3f08a612a65ba5b7f4fee9f2/generate/proto/generate.go#L69 ?

PR #29, opened 2019-10-13
Added templating for docker-compose

mihai93:


util/util.go
```diff
@@ -42,12 +42,20 @@ func TemplateFileAndOverwrite(fileDir string, fileName string, template *templat
...
 
 func TemplateFileIfDoesNotExist(fileDir string, fileName string, template *template.Template, data interface{}) {
-	fullFilePath := fmt.Sprintf("%v/%v", fileDir, fileName)
+	var fullFilePath string
+	
+	if fileDir == "" {
+		fullFilePath = fmt.Sprintf("%v", fileName)
+	} else {
+		fullFilePath = fmt.Sprintf("%v/%v", fileDir, fileName)
+	}
```
jb55: you can use `path.Join` when joining dirs to handle these cases

```go
fullFilePath = path.Join(fileDir, fileName)
```

PR #30, opened 2019-10-15
Adding react template

kb-commit:
This addresses couple of things:

1. Being able to develop react app with go template variables is not very feasible.. so this allows you to have files with regular extensions and files with `.tmpl` extnesions.. for example you can have `config.js` for running your react app while development and `config.js.tmpl` as templating file.. when skeleton is generated it will skip `config.js` and use `config.js.tmpl` to generate `config.js` file instead
2. Generates configurable frontend layout


config/react.go
```diff
@@ -0,0 +1,24 @@
+package config
+
+type reactApp struct {
```
kb-commit: these types are not exported to avoid namespace pollution

templates/react/package.json
```diff
@@ -1,11 +1,17 @@
 {
-  "name": "{{ .Name }}",
+  "name": "commit0",
```
kb-commit: this file is only used for local dev, see corresponding template file that is used in code generation

templates/react/src/App.js
```diff
@@ -1,26 +1,23 @@
...
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
-    </div>
+    <Layout>
+      <Router>
+        <Switch>
+          <Route path="/a">
```
kb-commit: these are just placeholders as we add more meaningful content

templates/react/src/config/index.js
```diff
@@ -0,0 +1,15 @@
+export default {
```
kb-commit: this file is only used for local dev, see corresponding template file that is used in code generation

templates/react/src/redux/reducers/object-reducer.js
```diff
@@ -0,0 +1,31 @@
+const initialState = {
```
kb-commit: create a generic reducer that we will be able to use for common redux stores..

templates/react/src/theme/index.js
```diff
@@ -0,0 +1,22 @@
+import { red } from '@material-ui/core/colors';
+import { createMuiTheme } from '@material-ui/core/styles';
+
+// A custom theme for this app
+const theme = createMuiTheme({
```
kb-commit: was thinking to templatize these variables, but thought might be an overkill

templator/templator.go
```diff
@@ -157,5 +164,24 @@ func getFileNames(box *packr.Box, dir string) []string {
...
 		return nil
 	})
-	return keys
+	return removeTmplDuplicates(keys)
+}
+
+func removeTmplDuplicates(keys []string) []string {
+	filteredKeys := []string{}
+	for _, key := range keys {
+		if !containsStr(keys, key+".tmpl") {
```
kb-commit: essentially if we have a `.tmpl` file with the same name, ignore the original file as its only used for local dev

PR #50, opened 2019-10-22
Cleaned up dockerfile, moved create test

bmonkman:


Dockerfile
```diff
@@ -39,24 +36,24 @@ RUN mkdir -p /tmp/protoc && \
...
 WORKDIR tmp/commit0
 COPY . .
 RUN make build-deps && make build && \
-  mv commit0 /usr/local/bin
-RUN upx --lzma /usr/local/bin/*
+  mv commit0 /usr/local/bin && \
+  upx --lzma /usr/local/bin/commit0
 
-FROM alpine:edge
+FROM alpine:3.10
```
bmonkman: Edge is the dev version, we should pin to a stable version.

Dockerfile
```diff
@@ -39,24 +36,24 @@ RUN mkdir -p /tmp/protoc && \
...
+RUN chmod +x /usr/local/bin/* && \
+  upx --lzma /usr/local/bin/*
+
 WORKDIR tmp/commit0
 COPY . .
 RUN make build-deps && make build && \
-  mv commit0 /usr/local/bin
-RUN upx --lzma /usr/local/bin/*
+  mv commit0 /usr/local/bin && \
+  upx --lzma /usr/local/bin/commit0
```
bmonkman: Separating this into two steps means most of the time we only need to re-compress the commit0 binary, not the other stuff that never changes.

.dockerignore
```diff
@@ -0,0 +1,7 @@
+Dockerfile
```
bmonkman: We should make sure to make this file as strict as possible, otherwise all this stuff gets included in `COPY . .` and busts the cache for all sorts of changes to various files.

PR #54, opened 2019-10-24
Clean up templator with a helper function, add readme (not filled in …

bmonkman:
…yet), run all template executions as goroutines


bmonkman:
Okay, made some more changes, should be ready now.

PR #65, opened 2019-11-01
#64 : Remove -l flag, move all configuration to yaml, a lot of refact…

bmonkman:
…oring
Mostly done, but I feel like the proto stuff might still need some love.


internal/generate/ci/generate_test.go
```diff
@@ -10,73 +10,20 @@ import (
 	"github.com/commitdev/commit0/internal/templator"
 )
 
-func TestGenerateJenkins(t *testing.T) {
```
Direside: I guess these cases are covered by the integration tests now anyway, plus the logic inside of the CI generator is much simpler now.

internal/generate/proto/generate.go
```diff
@@ -5,45 +5,36 @@ import (
...
+	serviceProtoDir := filepath.Join(idlPath, "proto", service.Name)
+	file := fmt.Sprintf("%s.proto", service.Name)
+	util.TemplateFileIfDoesNotExist(serviceProtoDir, file, t.ProtoServiceTemplate, wg, data)
+
+	GenerateProtoServiceLibs(service)
 }
 
-func GenerateProtoServiceLibs(config *config.Commit0Config) {
-	idlRoot := fmt.Sprintf("%s-idl", config.Name)
+func GenerateProtoServiceLibs(service config.Service) {
```
Pritesh-Patel: Before the idl repo produced would be a single repo that contained all the service idls & generated code split by folders, mainly for sharing protos (such as health) and for ease of use - is the intention to change this to multiple repos?

templates/proto/makefile.tmpl
```diff
@@ -31,36 +31,27 @@ deps-go:
...
 	go get -u github.com/golang/protobuf/protoc-gen-go
 
-generate: generate-grpc {{ if .Network.Web.Enabled }}generate-web{{- end}} {{ if .Network.Http.Enabled }}generate-http{{- end}}
+generate: generate-grpc {{ if .Service.Network.Web.Enabled }}generate-web{{- end}} {{ if .Service.Network.Http.Enabled }}generate-http{{- end}}
 
-generate-grpc:	
+generate-grpc:
 	mkdir -p gen/{{ $language }}
 	protoc ${PROTO_SOURCES} --{{ $language }}_out=plugins=grpc,paths=source_relative:proto ./proto/health/*.proto
-{{- range .Services}}
```
Pritesh-Patel: Related to the comment above, if this is removed how are the the service proto generation commands produced?

templates/proto/makefile.tmpl
```diff
@@ -31,36 +31,27 @@ deps-go:
...
 	cp -f -rv proto/Proto/* gen/web
 	rm -rf proto/proto proto/Proto
 {{- end}}
 
-{{- if .Network.Http.Enabled }}
+{{- if .Service.Network.Http.Enabled }}
 generate-http:
 	mkdir -p gen/go
 	protoc ${PROTO_SOURCES} --grpc-gateway_out=logtostderr=true,paths=source_relative:proto --swagger_out=logtostderr=true:proto ./proto/health/*.proto
-{{- range .Services}}
```
Pritesh-Patel: Same question for above

PR #67, opened 2019-11-06
Cleaned up terraform and fixed backends

bmonkman:
Refactored all templators to use a config holder so we can extend them in the future if we need more data.
Modified names of things in TF to reduce conflicts within the same AWS account.
Fixed S3 backend. Wasn't working at all because it can't live in a module, and can't contain string interpolation.

Still more to do to make everything fully repeatable, will probably do it in another ticket.

### DIFF
--- a/templates/kubernetes/terraform/environments/development/main.tf
+++ b/templates/kubernetes/terraform/environments/development/main.tf
@@ -1,15 +1,25 @@
+terraform {
+  backend "s3" {
+    bucket         = "project-{{ .Config.Name }}-terraform-state"
+    key            = "infrastructure/terraform/environments/development/main"
+    encrypt        = true
+    region         = "{{ .Config.Infrastructure.AWS.Region }}"
+    dynamodb_table = "terraform-state-locks"
+  }
+}
+
 # Instantiate the development environment
 module "development" {
   source      = "../../modules/environment"
   environment = "development"
 
   # Project configuration
-  project             = "{{ .Infrastructure.AWS.EKS.ClusterName }}"
-  region              = "{{ .Infrastructure.AWS.Region }}"
-  allowed_account_ids = ["{{ .Infrastructure.AWS.AccountId }}"]
+  project             = "{{ .Config.Infrastructure.AWS.EKS.ClusterName }}"
+  region              = "{{ .Config.Infrastructure.AWS.Region }}"
+  allowed_account_ids = ["{{ .Config.Infrastructure.AWS.AccountId }}"]
 
   # ECR configuration
-  ecr_repositories = ["{{ .Infrastructure.AWS.EKS.ClusterName }}"]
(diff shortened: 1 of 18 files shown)


templates/kubernetes/terraform/modules/vpc/main.tf
```diff
@@ -2,12 +2,12 @@ module "vpc" {
   source = "terraform-aws-modules/vpc/aws"
 
   name = "${var.project}-${var.environment}-vpc"
-  cidr = "10.10.0.0/16"
+  cidr = "10.20.0.0/16"
```
bmonkman: Didn't mean to commit this but I'll leave it for now. Should be fine.

templates/kubernetes/terraform/global/remote-state/main.tf
```diff
@@ -0,0 +1,24 @@
+provider "aws" {
```
bmonkman: To create the actual backend bucket and dynamo db. This will need to be run in a separate `tf apply` before the `tf init` of the rest of the infra. This will be an interesting piece to automate..

templates/kubernetes/terraform/environments/development/main.tf
```diff
@@ -1,15 +1,25 @@
+terraform {
```
bmonkman: The backend block can't be in a module and can't contain string interpolation so I moved it in here.

templates/kubernetes/terraform/global/remote-state/main.tf
ShahNewazKhan: We can move it to a seperate `.tf` file and use [terraform apply target](https://www.terraform.io/docs/commands/apply.html#target-resource) to apply specifically only that portion

bmonkman: That wouldn't be a great workflow going forward though, because you wouldn't be able to run plan/apply in the main dir.

ShahNewazKhan: How about using the `depends on` clause?

bmonkman: They have to be run separately, because one creates the global state and should never be run again, and the other one uses the global state.

ShahNewazKhan: Right, does it make sense to enclose the backend init in a shell script then or boto3?

ShahNewazKhan: That way we can both spin up and tear down the backend init based on `tf apply` and `tf destroy`

bmonkman: Gonna stick with this for now and we'll see how complex the process gets to see if it warrants adding helpers..

PR #72, opened 2019-11-09
added new react templates, github CI and terraform generators

zthomas:


templates/ci/github.tmpl
```diff
@@ -0,0 +1,29 @@
...
+
+    - run: npm test
+    
+    - run: npm run build
+
+    - name: upload build artifacts
+      uses: actions/upload-artifact@v1
+      with:
+        name: build
+        path: build
```
bmonkman: Not important now but you should make sure you have "add newline at end of file" turned in in your editor.

templates/react/.github/workflows/deploy.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Download build artifacts
+        uses: actions/download-artifact@v1
+        with:
+          name: build
+
+      # https://github.com/opspresso/action-s3-sync
+      - name: Publish to AWS S3
+        uses: opspresso/action-s3-sync@master
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
ShahNewazKhan: where is the secrets struct being hydrated?

bmonkman:
Feel free to just

templates/react/.github/workflows/deploy.yml
zthomas: Github Secrets. You just have to add them through their UI.

ShahNewazKhan: Right that's why I couldn't find it anywhere  :+1:

PR #75, opened 2019-11-11
Integrated terraform outputs in frontend code-gen

zthomas:
* updated frontend codebase to work with Cognito
* getting terraform outputs to be plugged back into our config (temporary solution)
* generating react environment variables with Cognito keys


internal/generate/generate_helper.go
```diff
@@ -45,6 +45,19 @@ func GenerateArtifactsHelper(t *templator.Templator, cfg *config.Commit0Config,
 		kubernetes.Generate(t, cfg, &wg, pathPrefix)
 	}
 
+	log.Println(aurora.Cyan(emoji.Sprintf("Generating Terraform")))
+	terraform.Generate(t, cfg, &wg, pathPrefix)
+	if cfg.Infrastructure.AWS.Cognito.Deploy {
+
```
zthomas: This is a temporary solution. We'll need to figure out a system to the scenarios where one part of the code generation requires some output from another step.

templates/terraform/main.tf
```diff
@@ -0,0 +1,81 @@
+provider "aws" {
+  region              = "${var.region}"
+}
+
```
zthomas: Let me know where we want to move these configs into later. I'm just putting everything into the top level `main.tf` for now

templates/terraform/variables.tf
```diff
@@ -0,0 +1,31 @@
...
+  default = "{{ .Config.Name }}"
+  description = "The name of the project, mostly for tagging"
+}
+
+variable "region" {
+  default = "{{ .Config.Infrastructure.AWS.Region }}"
+  description = "The AWS region"
+}
+
+# {{ if .Config.Infrastructure.AWS.Terraform.RemoteState }}
```
zthomas: I think this portion might be out of date after the last PR merge and we should be able to remove this.

templates/terraform/main.tf
```diff
@@ -0,0 +1,81 @@
...
+# {{ if .Config.Infrastructure.AWS.Terraform.RemoteState }}
+# Store remote state in S3
+resource "aws_s3_bucket" "terraform_remote_state" {
```
bmonkman: Can you remove this stuff please? It's already created by the bootstrap.

templates/terraform/main.tf
```diff
@@ -0,0 +1,81 @@
...
+    bucket         = "${var.remote_state_s3_bucket}"
+    key            = "infrastructure/terraform/shared"
+    encrypt        = true
+    region         = "${var.region}"
+    dynamodb_table = "${var.remote_state_dynamo_table}"
+  }
+}
+# {{- end}}
+# {{ if .Config.Infrastructure.AWS.Cognito.Deploy }}
+resource "aws_cognito_user_pool" "users" {
```
bmonkman: All this stuff already exists in the environment module, why is it duplicated again in main?

bmonkman: Or if it's not there already it should be there.

templates/terraform/variables.tf
bmonkman: I think you missed something about the refactor I did.. Let's chat?

internal/generate/generate_helper.go
bmonkman: Hm yeah, that's gonna get complicated..

internal/generate/generate_helper.go
```diff
@@ -45,6 +45,19 @@ func GenerateArtifactsHelper(t *templator.Templator, cfg *config.Commit0Config,
...
+		outputs := []string{
+			"cognito_pool_id",
+			"cognito_client_id",
+		}
+		outputValues := terraform.ExecuteWithOuput(cfg, pathPrefix, outputs)
+		cfg.Frontend.Env.CognitoPoolID = outputValues["cognito_pool_id"]
+		cfg.Frontend.Env.CognitoClientID = outputValues["cognito_client_id"]
```
bmonkman: I wonder if we should write these values back to the `commit0.yml` file..

templates/terraform/main.tf
bmonkman: This should be in the environment module with the other cognito stuff

internal/generate/generate_helper.go
zthomas: yup for sure, I think we definitely should do that

templates/terraform/main.tf
zthomas: kk removed.

templates/terraform/main.tf
zthomas: Okay just pushed out the updates

templates/terraform/modules/environment/main.tf
```diff
@@ -45,29 +45,45 @@ module "kube2iam" {
...
 # @TODO - Move this to a different file
 
-# {{ if .Config.Infrastructure.AWS.Cognito }}
-# ref: https://github.com/squidfunk/terraform-aws-cognito-auth#usage
-
-# data "aws_acm_certificate" "wildcard_cert" {
-#   domain   = "*.${var.public_dns_zone}"
-# }
-
-module "cognito-auth" {
```
bmonkman: This module wasn't working?

templates/terraform/modules/environment/variables.tf
```diff
@@ -32,14 +32,13 @@ variable "eks_worker_ami" {
...
 }
 
-# {{ if .Config.Infrastructure.AWS.Cognito }}
-variable "auth_namespace" {
-  default = "cognito_auth"
-}
-variable "auth_pool_name" {
+# {{ if .Config.Infrastructure.AWS.Cognito.Deploy }}
+variable "user_pool" {
+  default = "{{ .Config.Name }}"
```
bmonkman: This shouldn't be a default, you should provide this value as a parameter in `terraform/environments/<env>/main.tf`

templates/terraform/modules/environment/main.tf
zthomas: yah, had some issues. It was pretty complex and had a lot of moving parts. It never ran cleanly for me and also added some dependencies like `jq`

templates/terraform/modules/environment/variables.tf
zthomas: oh yup that sounds good.

templates/terraform/modules/environment/main.tf
bmonkman: 😢

templates/terraform/modules/environment/variables.tf
zthomas: Okay just pushed out the changes. So maybe a rule is that for variables that are same across all environments we can use default values otherwise it should not have a default but instead be defined in the environment specific input variables.

templates/terraform/modules/cognito/variables.tf
```diff
@@ -0,0 +1,7 @@
+variable "user_pool" {
+  description = "AWS Cognito pool name"
+} 
+variable "hostname" {
+  default = "{{ .Config.Frontend.Hostname }}"
```
bmonkman: This shouldn't be a default, it should be passed in from where the module is being used.

templates/terraform/modules/s3_hosting/main.tf
```diff
@@ -0,0 +1,103 @@
+resource "aws_s3_bucket" "www" {
+  // Our bucket's name is going to be the same as our site's domain name.
+  bucket = "${var.bucket_name}"
+  // Because we want our site to be available on the internet, we set this so
+  // anyone can read this bucket.
+  acl    = "public-read"
```
bmonkman: At some point this should be changed to private, with proper access set up from Cloudfront, we have plenty of examples of that. Not important for now though.

zthomas: yup true, I just wanted to be a bit easier for demoing right now. We can just use the s3 url to load it.

PR #77, opened 2019-11-13
Added prompts at project create, including lookups for aws credential…

bmonkman:
…s and account id


cmd/create.go
```diff
@@ -22,16 +29,63 @@ func Create(projectName string, outDir string, t *templator.Templator) string {
...
+		} else {
+			log.Fatalf(err.Error())
+		}
+	}
 
 	defaultProjConfig := defaultProjConfig(projectName)
 
+	defaultProjConfig.Infrastructure.AWS.Region = regionResult
+	if awsCaller != nil && awsCaller.Account != nil {
+		defaultProjConfig.Infrastructure.AWS.AccountID = *awsCaller.Account
```
zthomas: Nice!

cmd/create.go
```diff
@@ -22,16 +29,63 @@ func Create(projectName string, outDir string, t *templator.Templator) string {
 	rootDir := path.Join(outDir, projectName)
 	log.Println(aurora.Cyan(emoji.Sprintf(":tada: Creating project %s.", projectName)))
 	err := os.MkdirAll(rootDir, os.ModePerm)
-
```
zthomas: The command handlers should be kept slim and we should try to move all the logic deeper into some logical models. Love to plan out a clearer architecture for this project on Friday.

cmd/create_test.go
```diff
@@ -1,37 +1,30 @@
...
-	"github.com/commitdev/commit0/internal/util"
-	"github.com/gobuffalo/packr/v2"
 )
 
 func TestCreateWorks(t *testing.T) {
-	tmpdir, err := ioutil.TempDir("", "commit0-")
-	if err != nil {
-		t.Fatal(err)
-	}
+	// @TODO : Figure out a way to test this
```
ShahNewazKhan: Shane has some ci integration tests to check if the templates populated under `/tests`

internal/util/secrets/secrets.go
```diff
@@ -1,55 +1,96 @@
...
+		// Load the credentials file to look for profiles
+		credsFile := filepath.Join(usr.HomeDir, ".aws/credentials")
+		creds, err := ioutil.ReadFile(credsFile)
+		if err == nil {
+			// Get all profiles
+			re := regexp.MustCompile(`\[(.*)\]`)
+			profileMatches := re.FindAllStringSubmatch(string(creds), -1)
+			profiles := make([]string, len(profileMatches))
+			for i, p := range profileMatches {
+				profiles[i] = p[1]
```
ShahNewazKhan: should the `1` magic number be 0 to get the `default` profile?

ShahNewazKhan: nvm :man_facepalming:

cmd/create_test.go
bmonkman: Yeah, this test needs to be moved to a different level now because now this function triggers prompts.

internal/util/secrets/secrets.go
bmonkman: FindAllStringSubmatch returns a slice of tuples where index 0 is the full match and index 1 is the submatch

cmd/create.go
bmonkman: Yeah I left a todo there

Issue #59, opened 2019-10-25
Implement POC of the generator web UI


zthomas:
Implement POC of the generator web UI

Acceptance: 
- Implement all the features shown in the wireframes
- create a simple server to launch the UI and trigger commands

PR #85, opened 2019-11-30
Issue #59: Implement poc ui 

Direside:
Created a web-ui for creating projects.


internal/ui/src/App.css
```diff
@@ -0,0 +1,22 @@
...
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  font-size: calc(10px + 2vmin);
+  color: white;
+}
+
+.App-link {
+  color: #09d3ac;
+}
```
zthomas: Doesn't look like this file is still being used.

internal/ui/src/index.css
```diff
@@ -0,0 +1,13 @@
...
+    "Ubuntu", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue",
+    sans-serif;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
+
+code {
+  font-family: source-code-pro, Menlo, Monaco, Consolas, "Courier New",
+    monospace;
+}
```
zthomas: Is this being used?

internal/ui/src/App.css
Direside: Just the text-align, moved to index.css

internal/ui/src/index.css
Direside: Yes, it sets the font for the material ui components.

Issue #84, opened 2019-11-28
Propose a migration path to support fetching remote templates sources


zthomas:
AC: 
- A draft PR to show how our templating engine needs to be refactored.
- Address module config interface
- Address handling of templating logic for template modules
- Address how to handle template module context

PR #86, opened 2019-11-30
Issue #84 : proposal config syntax for package support

zthomas:


remote_templates/ci/.circleci/circleci.tmpl
```diff
@@ -0,0 +1,34 @@
+---
+delimiters
+  - "{{"
+  - "}}"
+---
```
zthomas: Optional frontmatter support. We don't necessarily need it right away but it does add a lot of flexibility.

remote_templates/ci/.github/workflows/github.tmpl
```diff
@@ -0,0 +1,52 @@
+name: Build and Deploy
```
zthomas: By default will be templated and written to the same path that it resides in the package. This could be further customized by frontmatter params if needed

remote_templates/commit0.example.yml
```diff
@@ -0,0 +1,24 @@
+name: hello-world
+
+# Context will populated automatically or could be added manually
+context: 
+  awsCognito: 
+    poolID: 123
+    clientID: ABC
+
```
zthomas: Note that the order of the package matter. Certain packages might require context generated from pervious packages. Eg. typically depending on infrastructure terraform outputs.

remote_templates/commit0.example.yml
```diff
@@ -0,0 +1,24 @@
...
+      apply: true
+
+  - source: './remote_templates/frontend'
+    params: 
+      ci: github
+
+  - source: './remote_templates/services'
+    # alternatively we can recursively support sub packges
+    packages: 
+      - './remote_templates/ci/go'
```
zthomas: Highlighting two different ways of dealing with including CI. 
1. as a package parameter that triggers a certain files to be included
2. as subpackages with the subpackage root path set to the parent's directory.

I prefer supporting sub-packages recursively, it encourages more reuse and separation of concerns.

modules/ci/.circleci/circleci.tmpl
```diff
@@ -0,0 +1,34 @@
+---
+delimiters
+  - "{{"
+  - "}}"
+---
```
zthomas: Optional frontmatter support. We don't need it right away but it does add a lot of flexibility.

modules/ci/.github/workflows/github.tmpl
```diff
@@ -0,0 +1,52 @@
+name: Build and Deploy
```
zthomas: By default files will be templated and written to the same path that it resides in the package. This could be further customized by frontmatter params if needed.

So this file will be templated and written to projectname/.github/workflows/github.yml

modules/commit0.example.yml
```diff
@@ -0,0 +1,25 @@
...
+      apply: true
+
+  - source: './remote_templates/frontend'
+    params: 
+      ci: github
+
+  - source: './remote_templates/services'
+    # alternatively we can recursively support sub packges
+    modules: 
+      - './remote_templates/ci/go'
```
zthomas: Highlighting two different ways of dealing with including CI.

as a package parameter that triggers a certain files to be included
as subpackages with the subpackage root path set to the parent's directory.
I prefer supporting sub-packages recursively, it encourages more reuse and separation of concerns.

modules/ci/.circleci/circleci.tmpl
bmonkman: Why would these be specified here if they are already in the module yaml? To be able to override on a per-file basis?

modules/ci/commit0.module.yml
```diff
@@ -0,0 +1,24 @@
...
+  strictMode: true # will only parse files that includes the .tmpl.* extension, otherwise it will copy file 
+  delimiters:
+    - "<%"
+    - "%>"
+  destination: "{{.Metadata.Directory}}/{{.Metadata.Filename}}"
+
+# parameters required 
+# populates .Params variables
+params: 
+  - name: system
```
bmonkman: What would this affect?

modules/commit0.example.yml
```diff
@@ -0,0 +1,25 @@
...
+context: 
+  awsCognito: 
+    poolID: 123
+    clientID: ABC
+
+modules: 
+  # module can be in any format the go-getter supports (path, github, url, etc.)
+  - source: './remote_templates/infrastructure'
+    params: 
+      apply: true
```
bmonkman: Something like whether or not to apply should probably be controlled by the user at runtime.

modules/ci/.circleci/circleci.tmpl
zthomas: yup all front-matter variables will be for per file overrides. We totally don't need this feature right away but I just wanted to include it as part of the draft.

modules/ci/commit0.module.yml
zthomas: so `params` would essentially be the form builder or CLI prompt configs. This is inspired by https://yeoman.io/authoring/user-interactions.html

modules/commit0.example.yml
zthomas: Yup that makes sense.

modules/ci/commit0.module.yml
```diff
@@ -0,0 +1,24 @@
...
+  title: "CI templates"
+  description: ""
+  author: ""
+  icon: ""
+  thumbnail: ""
+  readme: "" # defaults to README.md
+
+# Template variables to populate, these could be overwritten by the file spefic frontmatter variables
+templates: 
+  strictMode: true # will only parse files that includes the .tmpl.* extension, otherwise it will copy file
```
Direside: Not sure on the naming here, also wouldn't we just parse all the files listed by default, or maybe do a .commitignore file to tell it to not parse certain files that aren't needed. For example maybe the README.md in the folder could be ignored, but the rest copied over.

zthomas: Sorry I didn't see your message earlier. Are you talking just about the `strictMode`? Yah we probably don't need this just yet. I thought of it as a way to improve performance. For a really sparsely templated project this might be a speed up.

PR #88, opened 2019-12-13
Updated templates and code to reflect newest best state from aplayr a…

bmonkman:
…nd winedirect

Updated EKS AMIs
Added uuid generation function for templates
Created multiple remote states for multi env
Added k8s external_dns support
Productionized S3/Cloudfront setup by adding proper access control, ACM, R53
Added support for multiple buckets
Added ASG min for EKS


wdbenjamin:
Looks good to me Bill! Thank you for bringing a number of these changes across from the other repo!

PR #102, opened 2020-03-20
Added CircleCI pipeline to test on branch and build and push on master.

Direside:


trevoro:
Hey not sure if we want to add this into this PR or a separate issue but it's helpful to identify the build from the version. For example, the build could be based on the git sha and not the hardcoded version. The tag could still be the version I suppose, but it's helpful to be able to know which commit each build came from if someone forgets to increment the semver in the `Makefile`

Example Makefile would be something like this:

```
BUILD ?=$(shell git rev-parse --short HEAD)
BUILD_ARGS=-ldflags "-X main.appVersion=${VERSION} -X main.appBuild=${BUILD}"

correct_target_name:
	CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo ${BUILD_ARGS}
```

Then in the app / main entrypoint / wherever we do this:

```
package main
...
...
var (
    appVersion = "SNAPSHOT" // default value
    appBuild = "SNAPSHOT      // default value
)
```

When built the variables are injected from the build command and default to these values if not provided.

Direside:
Yeah we should come up with a versioning scheme at some point. This PR uses the `version-tag` orb which will label the docker images as the CircleCI build number + the short git hash. 

E.g. `5-ddc70ff`

The problem with this approach is that it's created at build time so we'd need to add some more steps to the build to make it be able to print that version tag when you run `commit0 version`. 

The other approach is to update the number in `cmd/version.go` and update it with each PR, but that requires us to remember to edit that every time we make a code change.

trevoro:
The only step you need to add to the build are the ones I outlined in the Makefile example earlier plus the global version variable in the code. I'll just add it later after this PR.

Direside:
Sounds good. I found a way to make it work by just updating the file using sed for now. I'll let you make the required changes when you come up with a proper versioning plan.

Current behaviour:
```sh
docker run commitdev/commit0:9-c1d4069 version
9-c1d4069
```

PR #106, opened 2020-03-22
cleaned up unused prompts

zthomas:
- removed unused prompts, they have been moved into the modules
- updated the sample `commit0.yml` config to point to a commitdev org repo instead of personal repo


trevoro:
Looks fine to me. Any idea why it's not building properly?

PR #121, opened 2020-05-16
Added command to check your local installs.

Direside:


cmd/check.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+
+		// Store and errors from the commands we run.
+		errors := []commandError{}
+
+		fmt.Println("Checking Zero Requirements...")
+		for _, r := range required {
+			fmt.Printf("%s", r.name)
+			out, err := exec.Command(r.command, r.args...).CombinedOutput()
+			if err != nil {
```
bmonkman, line 204: Can we make a differentiation here between "the command doesn't exist" and "executing this command returned a non-zero exit code"?

cmd/check.go
```diff
@@ -0,0 +1,246 @@
...
+
+			fmt.Fprintf(w, "\n %s\t%s\t%s\t", "Command", "Error", "Info")
+			fmt.Fprintf(w, "\n %s\t%s\t%s\t", "---------", "---------", "---------")
+
+			for _, e := range errors {
+				fmt.Fprintf(w, "\n%s\t%s\t%s\t", e.Command, e.ErrorText, e.Suggestion)
+			}
+		}
+		fmt.Println()
+	},
```
bmonkman, line 245: Should return zero if the check succeeded, non-zero if it failed.

cmd/check.go
```diff
@@ -0,0 +1,246 @@
...
+						return ver, err
+					}
+					patch, err := strconv.ParseInt(m[3], 0, 64)
+					if err != nil {
+						return ver, err
+					}
+
+					ver = fmt.Sprintf("%d.%d.%d", major, minor, patch)
+
+					if major < 1 || (major == 1 && minor < 16) {
```
bmonkman, line 75: Can we push these version numbers up to config? Even if it's just hard-coded config for now.

cmd/check.go
```diff
@@ -0,0 +1,246 @@
...
+	Use:   "check",
+	Short: "Print the check number of commit0",
+	Run: func(cmd *cobra.Command, args []string) {
+		// Add any new requirements to this slice.
+		required := []requirement{
+			{
+				name:    "AWS CLI\t\t",
+				command: "aws",
+				args:    []string{"--version"},
+				docsURL: "",
```
Direside, line 55: Should add the AWS CLI install URL here also.

cmd/check.go
```diff
@@ -0,0 +1,246 @@
...
+
+func init() {
+	rootCmd.AddCommand(checkCmd)
+}
+
+type requirement struct {
+	name    string
+	command string
+	args    []string
+	docsURL string
```
bmonkman, line 23: At some point we could have our own documentation string here that could have an OS-specific description of how to install this tool. (hopefully just a single command)

cmd/check.go
Direside: Yeah, I'll investigate this some more. I was able to infer based on the output of the command if it's missing or not as the error is consistent. We would need to check cross-platform to make sure it returns a similar error. There may also be another way to run commands that is more verbose.

E.g.
```zsh
 Command         Error                                            Info                                                     
 ---------       ---------                                        ---------                                                
kubectl version  exit status 1                                    https://kubernetes.io/docs/tasks/tools/install-kubectl/  
jq2 --version    exec: "jq2": executable file not found in $PATH  https://stedolan.github.io/jq/download/                  
git              Zero requires git version 2.12 or greater.       https://git-scm.com/book/en/v2/Getting-Started-Installing-Git% 
```

cmd/check.go
Direside: Can we reliably count on all the tools we are checking adhering to some version of semver? If so we could have this as part of the config or even use a semver library to do these calculations.

Direside: My initial thought was to simply have each command use a regex and a semver minimum version as part of the config, then push all the logic to a shared function. But I was concerned that perhaps some tools wouldn't fix that. Given all the current tools allow that I might just go back to it.

bmonkman: Looks good. If we introduce a dependency that doesn't use semver we can reassess then.

PR #123, opened 2020-05-22
First pass at renaming and cleaning up old stuff

bmonkman:


OLD-README.md
```diff
@@ -83,18 +83,18 @@ This is a guide on how to configure your project manually with a single file `co
...
-context: 
+context:
   cognitoPoolID: xxx
-modules: 
-  - source: "github.com/commitdev/commit0-aws-eks-stack"	
+modules:
+  - source: "github.com/commitdev/commit0-aws-eks-stack"
     output: "infrastructure"
-  - source: "github.com/zthomas/react-mui-kit"	
+  - source: "github.com/zthomas/react-mui-kit"
```
Direside, line 95: Did we want to replace this with the other Frontend stack?

bmonkman: That's the old readme that can probably be deleted but I figured we should keep around for a little bit for reference.

PR #141, opened 2020-06-01
ZERO126 adding repo creation and initial commit

cjhawkins:


cmd/create-repos.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+		}
+	}
+
+	fmt.Printf("Repository successfully initialized for module: %s\n", moduleDir)
+
+	return nil
+}
+
+func createReadme(moduleDir string) error {
```
bmonkman, line 204: There will already be a readme in the repository, this shouldn't be necessary.

bmonkman:
Mind moving this to `internal`? Maybe `internal/vcs` or `internal/repositories`?

Issue #135, opened 2020-05-29
Init Command - update location of credentials file, make sure functionality exists (currently missing git support), change structure of the file to support multiple projects


bmonkman:
From file format document:

credentials.yml format:
This file would be created by init and contain credentials to access various services, keyed by project name.
`~/.zero/credentials.yml`
```
my-project:
- aws:
  - accessKeyId: AKIAABCD
  - secretAccessKey: ZXCV
- circleci:
  - apiKey: "1234"
- github:
  - accessToken: "0987"
another-project:
- github:
  - accessToken: "654"
```

Issue #134, opened 2020-05-29
Init Command - Prompt user to create repos and check in code


bmonkman:
From UX doc:

(Git stuff - required for us to be able to create repos and commit projects - results will be stored in zero-project.yml)

covered by #143
>~? Should the created projects be checked into github automatically? (Y/n: y~
~? What’s the root of the github org to create repositories in? : github.com/my-org/~
~? Github Personal Access Token with access to the above organization : ABCD1234~

Outstanding
```
? What do you want to call the “zero-aws-eks-stack” project? (infrastructure): infrastructure
? What do you want to call the “zero-deployable-backend” project? (backend): backend-service
? What do you want to call the “zero-deployable-react-frontend” project? (frontend): frontend
```
See the file structure document for some more details. 

Actual creation will be done during the create command.
Related to #126

Issue #127, opened 2020-05-26
Init Command - Pull in modules and load config during the init phase - deduplicate config options for prompting


bmonkman:

davidcheung:
covered by #143

PR #143, opened 2020-06-01
dedupe

davidcheung:
https://github.com/commitdev/zero/issues/134
- [x]  Should the created projects be checked into github automatically? (Y/n): y
- [x] What’s the root of the github org to create repositories in? : github.com/my-org/
- [x] Github Personal Access Token with access to the above organization : ABCD1234

https://github.com/commitdev/zero/issues/127
- [x] Dedupe based on elimination(passed context and remove by key) 

https://github.com/commitdev/zero/issues/135
- [x] creating ~/.zero/credentials.yml 
- [x] reading ~/.zero/credentials.yml 
- [x] writing basic project config to ~/.zero/credentials.yml

 
Outstanding 
- [ ] per module ask for directory name to use [backend/infrastructure/frontned]
- [ ] tests for any the prompts

### DIFF
--- a/internal/config/global_config.go
+++ b/internal/config/global_config.go
@@ -1,0 +1,123 @@
+package config
+
+import (
+	"bytes"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/user"
+	"path"
+
+	"github.com/commitdev/zero/configs"
+	"github.com/commitdev/zero/pkg/util/exit"
+	yaml "gopkg.in/yaml.v2"
+)
+
+var GetCredentialsPath = getCredentialsPath
+
+type ProjectCredentials map[string]ProjectCredential
+
+type ProjectCredential struct {
+	ProjectName            string `yaml:"-"`
+	AWSResourceConfig      `yaml:"aws,omitempty"`
+	GithubResourceConfig   `yaml:"github,omitempty"`
+	CircleCiResourceConfig `yaml:"circleci,omitempty"`
+}
+
+type AWSResourceConfig struct {
+	AccessKeyId     string `yaml:"accessKeyId,omitempty"`
+	SecretAccessKey string `yaml:"secretAccessKey,omitempty"`
+}
+type GithubResourceConfig struct {
+	AccessToken string `yaml:"accessToken,omitempty"`
+}
+type CircleCiResourceConfig struct {
+	ApiKey string `yaml:"apiKey,omitempty"`
+}
+
+func (p ProjectCredentials) Unmarshal(data []byte) error {
+	if len(data) == 0 {
+		return nil
+	}
+	err := yaml.NewDecoder(bytes.NewReader(data)).Decode(p)
+	if err != nil {
+		return err
+	}
+	for k, v := range p {
+		v.ProjectName = k
+		p[k] = v
+	}
+	return nil
+}
+
+func LoadUserCredentials() ProjectCredentials {
+	data := readOrCreateUserCredentialsFile()
+
+	projects := ProjectCredentials{}
+
+	err := projects.Unmarshal(data)
+
+	if err != nil {
+		exit.Fatal("Failed to parse configuration: %v", err)
+	}
+	return projects
+}
+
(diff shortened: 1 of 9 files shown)


bmonkman:
Based on @cjhawkins 's PR regarding github it seems like we may need to switch around what we ask for here since we need to know if it's a personal repo or organizational, but we can handle that in another PR.

internal/config/global_config.go
```diff
@@ -0,0 +1,117 @@
...
+	"os/user"
+	"path"
+
+	"github.com/commitdev/zero/configs"
+	yaml "gopkg.in/yaml.v2"
+)
+
+var GetCredentialsPath = getCredentialsPath
+
+type Projects map[string]*Project
```
bmonkman, line 17: Is there a reason for these to be pointers to structs instead of just structs?

bmonkman: Also can we call this something more specific? `Credentials` and `ProjectCredentials`?

internal/context/init.go
```diff
@@ -56,18 +58,77 @@ func promptAllModules(projectConfig *config.ZeroProjectConfig) {
...
+
+func promptGithubRootOrg(projectConfig *config.ZeroProjectConfig) {
+	providerPrompt := promptui.Prompt{
+		Label:     "What's the root of the github org to create repositories in?",
+		Default:   "github.com/",
+		AllowEdit: true,
+	}
+	providerResult, err := providerPrompt.Run()
+	if err != nil {
+		panic(err)
```
bmonkman, line 87: Let's use `exit.Fatal` instead of panicing

internal/context/init.go
```diff
@@ -32,18 +32,20 @@ func Init(projectName string, outDir string) *config.ZeroProjectConfig {
 
 	projectConfig := defaultProjConfig(projectName)
 	promptProjectName(projectName, &projectConfig)
+	promptPushRepoUpstream(&projectConfig)
```
bmonkman, line 35: Not a big fan of passing pointers here. It would be more clear to write statements like `projectConfig.Context["ShouldPushRepoUpstream"] = promptPushRepoUpstream(projectConfig)`

internal/config/global_config_test.go
```diff
@@ -0,0 +1,103 @@
+package config_test
```
bmonkman, line 1: Nice tests!

internal/config/global_config.go
davidcheung: yeah, I need to persist the changes I make to the Project,  the map[string]Project doesn't get the updated value otherwise

bmonkman: Only because `Save` is a pointer receiver. It could also be called like `config.Save(projects)` instead of `projects.Save()` There are a few things in this PR which look like they are structured in an OOP way which is not how things are typically done in Go.

internal/config/global_config_test.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	err = ioutil.WriteFile(to, bytesRead, 0644)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+func TestReadOrCreateUserCredentialsFile(t *testing.T) {
```
bmonkman, line 42: These should really be testing the API of the package rather than the internals. By calling a function like `ReadOrCreateUserCredentialsFile` it means that function now has to be exported, and I would think the only exported functions would be `LoadUserCredentials` and `Save`

davidcheung: Make sense, should I create another file for tests like this under the `config` package? so I can keep them private and still test them?

bmonkman: Maybe, though the general practice is try to make your tests in the _test package, then test the exported stuff and not the internals so that your test is coupled to the "API" which should be stable and not the internals which could change. And if you get to the point where the "API" is masking a massive amount of code, try to split it up into smaller packages which can be easily tested in isolation.

PR #145, opened 2020-06-02
[apply] execute make in each project

### LINKED ISSUE #132: Apply Command - Execute make in each project
Execute make in each directory created by the create command. Initially just run them in series and print all the output to stdout. We'll figure out how to make a better experience.

thdaraujo:
Apply Command - Make Projects [closes #132]

WIP


internal/config/projectconfig/project_config.go
```diff
@@ -9,10 +9,10 @@ import (
 )
 
 type ZeroProjectConfig struct {
-	Name           string
-	Infrastructure Infrastructure // TODO simplify and flatten / rename?
-	Context        map[string]string
-	Modules        []string
+	Name           string            `yaml:"name"`
```
bmonkman, line 12: This will likely cause conflicts with @dtoki 's stuff but we can figure it out.

internal/config/projectconfig/apply.go
```diff
@@ -0,0 +1,43 @@
+package projectconfig
```
bmonkman, line 1: I don't think this is the right place for this. Maybe move to an `internal/apply` package?

internal/context/apply.go
```diff
@@ -0,0 +1,65 @@
+package context
```
bmonkman, line 1: Move to an `internal/apply` package?

internal/context/apply.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/manifoldco/promptui"
+)
+
+// Apply will load the context/configuration to be used by the apply command
+func Apply(applyEnvironments []string, applyConfigPath string) *projectconfig.ZeroProjectConfig {
+	if len(applyEnvironments) == 0 {
+		fmt.Println(`Choose the environments to apply. This will create infrastructure, CI pipelines, etc.
+At this point, real things will be generated that may cost money!
+Only a single environment may be suitable for an initial test, but for a real system we suggest setting up both staging and production environments.`)
+		applyEnvironments = promptEnvironments()
```
bmonkman, line 19: This won't affect the value of `applyEnvironments` in the calling code, but it might be better to merge this and the stuff in the other file into the same package anyway.

internal/context/init.go
```diff
@@ -63,11 +63,11 @@ func Init(projectName string, outDir string) *projectconfig.ZeroProjectConfig {
 }
 
 // loadAllModules takes a list of module sources, downloads those modules, and parses their config
-func loadAllModules(moduleSources []string) map[string]moduleconfig.ModuleConfig {
+func loadAllModules(moduleSources projectconfig.Modules) map[string]moduleconfig.ModuleConfig {
```
bmonkman, line 66: This shouldn't accept projectconfig.Modules, just a slice of string. At this point there is no project config and all we have are some repo paths.

internal/config/projectconfig/apply.go
```diff
@@ -0,0 +1,43 @@
...
+	flog.Infof("Infrastructure executor: %s", "Terraform")
+
+	// other details...
+
+	return makeAll(dir, projectContext, applyEnvironments)
+}
+
+func makeAll(dir string, projectContext *ZeroProjectConfig, applyEnvironments []string) []string {
+	environmentArg := fmt.Sprintf("ENVIRONMENT=%s", strings.Join(applyEnvironments, ","))
+	envars := []string{environmentArg}
```
bmonkman, line 30: Could you also add the parameters from the project config as env vars?

internal/context/init.go
thdaraujo: changed it back to a string

bmonkman: Thanks, could you remove the changes to the registry section as well?

internal/config/projectconfig/apply.go
thdaraujo: done

internal/util/util.go
```diff
@@ -101,3 +102,14 @@ func ExecuteCommandOutput(cmd *exec.Cmd, pathPrefix string, envars []string) str
 	}
 	return string(out)
 }
+
+// AppendProjectEnvToCmdEnv will add all the keys and values from envMap
+// into envList as key-value pair strings (e.g.: "key=value")
+func AppendProjectEnvToCmdEnv(envMap map[string]string, envList []string) []string {
```
bmonkman, line 108: This will conflict with a change in David's PR since he also moved this function. But I agree it should be here in util.

PR #146, opened 2020-06-03
Reorganize config code by moving different config types into better-n…

bmonkman:
…amed packages


internal/module/module.go
```diff
@@ -49,7 +50,7 @@ func NewTemplateModule(moduleCfg config.ModuleInstance) (*TemplateModule, error)
 	}
 
 	configPath := path.Join(sourcePath, "zero-module.yml")
```
davidcheung, line 52: can perhaps move this to constants too

PR #153, opened 2020-06-09
Removed the required project name when running the command to make it…

Direside:
… simpler to run.


Direside:
It seemed kind of annoying to me that it requires you to have the project name when you run the command as an argument, but then immediately lets you change it.

I think that it's a better experience if we just let people type the project name as part of the command running, requiring positional command line arguments can be annoying for people if they don't know how to use the tool.

PR #154, opened 2020-06-09
support module-config default / value

davidcheung:
- also adding support for project conditional prompts

### DIFF
--- a/internal/context/init.go
+++ b/internal/context/init.go
@@ -24,7 +24,8 @@ type Registry map[string][]string
 // Create cloud provider context
 func Init(outDir string) *projectconfig.ZeroProjectConfig {
 	projectConfig := defaultProjConfig()
-	projectConfig.Name = promptProjectName()
+
+	projectConfig.Name = getProjectNamePrompt().GetParam(projectConfig.Parameters)
 
 	rootDir := path.Join(outDir, projectConfig.Name)
 	flog.Infof(":tada: Creating project")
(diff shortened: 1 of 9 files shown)


internal/context/init.go
```diff
@@ -81,77 +87,49 @@ func promptAllModules(modules map[string]moduleconfig.ModuleConfig) map[string]s
...
-		Label:     "Project Name",
-		Default:   projectName,
-		AllowEdit: false,
-	}
-	result, err := providerPrompt.Run()
-	if err != nil {
-		exit.Fatal("Prompt failed %v\n", err)
+func getProjectPrompts(projectName string) map[string]PromptHandler {
+	return map[string]PromptHandler{
+		"projectName": {
```
bmonkman, line 100: Since it was supplied on the command line we probably don't need to prompt for name. Or maybe we make it optional on the command line and then conditionally prompt here?

internal/context/init.go
```diff
@@ -81,77 +87,49 @@ func promptAllModules(modules map[string]moduleconfig.ModuleConfig) map[string]s
...
+			},
+			NoCondition,
+		},
+		"GithubRootOrg": {
+			moduleconfig.Parameter{
+				Field:   "GithubRootOrg",
+				Label:   "What's the root of the github org to create repositories in?",
+				Default: "github.com/",
+			},
+			KeyMatchCondition("ShouldPushRepoUpstream", "y"),
```
bmonkman, line 122: Nice implementation

internal/context/prompts.go
```diff
@@ -0,0 +1,129 @@
...
+func appendProjectEnvToCmdEnv(envMap map[string]string, envList []string) []string {
+	for key, val := range envMap {
+		if val != "" {
+			envList = append(envList, fmt.Sprintf("%s=%s", key, val))
+		}
+	}
+	return envList
+}
+
+// TODO : Use this function signature instead
```
bmonkman, line 111: You can remove this todo

internal/context/init.go
```diff
@@ -34,16 +35,21 @@ func Init(projectName string, outDir string) *projectconfig.ZeroProjectConfig {
...
-	// fmt.Println(&projectConfig)
-	// s := project.GetSecrets(rootDir)
-	// fillProviderDetails(&projectConfig, s)
-	// fmt.Println(&projectConfig)
+
+	prompts := getProjectPrompts(projectName)
+
+	projectConfig.Name = prompts["projectName"].GetParam(projectConfig.Parameters)
+
+	projectConfig.Parameters["ShouldPushRepoUpstream"] = prompts["ShouldPushRepoUpstream"].GetParam(projectConfig.Parameters)
```
bmonkman, line 43: You don't need to do this as part of this PR but shouldn't these be part of the projectConfig struct rather than parameters? Parameters are passed into project templates, but these values we will be using for zero itself during the create phase.

internal/context/init.go
bmonkman: I'll do this as part of my ticket.

internal/context/init.go
bmonkman: I'll do this as part of my ticket.

internal/context/prompts.go
```diff
@@ -0,0 +1,129 @@
...
+	return true
+}
+func KeyMatchCondition(key string, value string) func(map[string]string) bool {
+	return func(param map[string]string) bool {
+		return param[key] == value
+	}
+}
+
+// TODO: validation / allow prompt retry ...etc
+func (p PromptHandler) GetParam(projectParams map[string]string) string {
```
bmonkman, line 30: Seems like this would be a good one to have a test for, exercising the different types.

davidcheung: i did, i aliased my `git a` to `git add . -u` and forgot to index the untracked file 😅

internal/context/init.go
davidcheung: https://github.com/commitdev/zero/pull/153 yeah we removed the projectName as a positional argument altogether just a little earlier, ill need to rebase with that change

PR #162, opened 2020-06-12
ZER0155 update repo init to detect owner type

cjhawkins:


internal/vcs/create-git-repos.go
```diff
@@ -5,34 +5,71 @@ import (
...
-// 1. create.go will handle parsing the repo names and github credentials and pass them into initializeRepostiories().
-// 2. if organizationName is set, create an org owned repo.  if not, create a personal repo.
+// InitializeRepository Creates and initializes a github repository for the given url
+// repositoryUrl is expected to be in the format "github.com/{ownerName}/{repositoryName}"
+func InitializeRepository(repositoryUrl string, githubApiKey string) {
 
-// takes a list of subdirectories containing modules to create a repository and do initial commit for
-func InitializeRepositories(moduleDirs []string, remoteRepository string, organizationName string, githubApiKey string) {
+	ownerName, repositoryName, pErr := parseRepositoryUrl(repositoryUrl)
+	if pErr != nil {
```
bmonkman, line 16: Any reason to not just have `err` and reuse it?

cjhawkins: I like to keep errors scoped as close to the call as possible, but its not really needed here.  I can change this to reuse a single err.

internal/vcs/create-git-repos.go
```diff
@@ -96,51 +124,42 @@ type organizationQueryResponse struct {
...
-func doInitialCommit(moduleDir string, remoteRepository string) error {
-	fmt.Printf("Initializing repository for module: %s\n", moduleDir)
+// doInitialCommit runs the git commands that initialize and do the first commit to a repository.
+func doInitialCommit(ownerName string, repositoryName string) error {
+	fmt.Printf("Initializing repository for module: %s\n", repositoryName)
 
-	remoteOrigin := fmt.Sprintf("%s/%s.git", remoteRepository, moduleDir)
-	commands := []InitialCommands{
+	remoteOrigin := fmt.Sprintf("git@github.com:%s/%s.git", ownerName, repositoryName)
+	fmt.Printf("remote origin: %s\n", remoteOrigin)
```
bmonkman, line 162: Probably not necessary to print

internal/vcs/create-git-repos.go
```diff
@@ -5,34 +5,72 @@ import (
...
+
+	segments := strings.Split(repositoryUrl, "/")
+	if len(segments) != 3 {
+		return "","", fmt.Errorf("invalid repository url.  expected format \"github.com/{ownerName}/{repositoryName}\"")
+	}
+
+	ownerName := segments[1]
+	repositoryName := segments[2]
+
+	fmt.Printf("found owner %s, repository %s\n", ownerName, repositoryName)
```
bmonkman, line 71: not necessary to print

internal/vcs/create-git-repos.go
```diff
@@ -84,8 +98,23 @@ const createOrganizationRepositoryMutation = `mutation ($repoName: String!, $rep
...
+func createRepository(request *graphql.Request, repositoryName string) error {
+	fmt.Printf("Creating repository for module: %s\n", repositoryName)
+
+	c := graphql.NewClient("https://api.github.com/graphql")
+	ctx := context.Background()
+	if err := c.Run(ctx, request, nil); err != nil {
+		return err
+	}
+
+	fmt.Printf("Repository successfully created for module: %s\n", repositoryName)
```
bmonkman, line 111: Just the above print should be enough

internal/vcs/create-git-repos.go
bmonkman: Actually I think all the status prints in this probably aren't very useful to the user. Just one per repo saying we are creating a repo should be enough.

PR #164, opened 2020-06-16
Global cred prompt

davidcheung:


internal/context/prompts.go
```diff
@@ -147,3 +149,43 @@ func PromptModuleParams(moduleConfig moduleconfig.ModuleConfig, parameters map[s
...
+		for _, prompt := range prompts {
+			vendorPromptValues[prompt.Field] = prompt.GetParam(map[string]string{})
+		}
+		promptsValues[vendor] = vendorPromptValues
+	}
+
+	// FIXME: what is a good way to dynamically modify partial data of a struct
+	// current just marashing to yaml, then unmarshaling into the base struct
+	yamlContent, _ := yaml.Marshal(promptsValues)
+	yaml.Unmarshal(yamlContent, &credentials)
```
bmonkman, line 169: Hm yeah that's not ideal..

internal/context/prompts.go
```diff
@@ -147,3 +149,43 @@ func PromptModuleParams(moduleConfig moduleconfig.ModuleConfig, parameters map[s
...
+
+	globalconfig.Save(credentials)
```
bmonkman, line 171: The name of this function doesn't make it very clear that it's actually going to save the file. Maybe rename or just return the credentials and let the caller save it?

internal/context/prompts.go
bmonkman: Probably okay for now though..

PR #165, opened 2020-06-17
Marshal zero-project.yaml file

### LINKED ISSUE #130: Init command - Template the prompted user input into zero-project.yml
Currently it only templates in the name 

we will need prompted values in there for `create`

dtoki:
**Pending:**
~Util function to properly indent the modules.~
~Asserts statements on `TestGetProjectFileContent`~

Closes #130

### DIFF
--- a/internal/config/projectconfig/init.go
+++ b/internal/config/projectconfig/init.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/commitdev/zero/internal/constants"
 	"github.com/commitdev/zero/internal/util"
-	"github.com/commitdev/zero/pkg/util/exit"
 	"gopkg.in/yaml.v2"
 )
 
@@ -29,16 +28,19 @@ func SetRootDir(dir string) {
 	RootDir = dir
 }
 
-func CreateProjectConfigFile(dir string, projectName string, projectContext *ZeroProjectConfig) {
+// CreateProjectConfigFile extracts the required content for zero project config file then write to disk.
+func CreateProjectConfigFile(dir string, projectName string, projectContext *ZeroProjectConfig) error {
 	content, err := getProjectFileContent(*projectContext)
 	if err != nil {
-		exit.Fatal(fmt.Sprintf("Failed extracting the file config content %s", constants.ZeroProjectYml))
+		return err
 	}
 
 	writeErr := ioutil.WriteFile(path.Join(dir, projectName, constants.ZeroProjectYml), []byte(content), 0644)
 	if writeErr != nil {
-		exit.Fatal(fmt.Sprintf("Failed to create config file %s", constants.ZeroProjectYml))
+		return err
 	}
+
+	return nil
 }
 
 func getProjectFileContent(projectConfig ZeroProjectConfig) (string, error) {
--- a/internal/config/projectconfig/init.go
+++ b/internal/config/projectconfig/init.go
@@ -1,43 +1,28 @@
 package projectconfig
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
+	"text/template"
 
 	"github.com/commitdev/zero/internal/constants"
+	"github.com/commitdev/zero/internal/util"
 	"github.com/commitdev/zero/pkg/util/exit"
 	"gopkg.in/yaml.v2"
 )
 
-const exampleConfig = `name: %s
+// {{  .ShouldPushRepositories | printf "%q"}}
+const exampleConfig = `
+# Templated zero-project.yml file
+name: {{.Name}}
 
-# Context is normally populated automatically but could be used to inject global params
-context:
+shouldPushRepositories: {{.ShouldPushRepositories | printf "%v"}}
 
-# module can be in any format the go-getter supports (path, github, url, etc.)
-# supports https://github.com/hashicorp/go-getter#url-format
-# Example:
-# - repo: "../development/modules/ci"
-# - dir: "github-actions"
 modules:
-	aws-eks-stack:
-		parameters:
-			repoName: infrastructure
-			region: us-east-1
-			accountId: 12345
-			productionHost: something.com
-		files:
-			dir: infrastructure
-			repo: https://github.com/myorg/infrastructure
-	some-other-module:
-		parameters:
-			repoName: api
-		files:
-			dir: api
-			repo: https://github.com/myorg/api
-
-
+{{.Modules}}
 `
 
 var RootDir = "./"
@@ -58,4 +43,34 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
 	if writeErr != nil {
 		exit.Fatal(fmt.Sprintf("Failed to create config file %s", constants.ZeroProjectYml))
 	}
+}
+
+func GetProjectFileContent(projectConfig ZeroProjectConfig) string {
+	var tpl bytes.Buffer
+	tmpl := template.New("sa")
+	tmpl, err := tmpl.Parse(exampleConfig)
+	if err != nil {
+		exit.Fatal(fmt.Sprintf("Failed to parse the sample Zero module config file %s", constants.ZeroProjectYml))
+	}
+
+	type tempProjectConfig struct {
+		Name                   string
+		ShouldPushRepositories bool
+		Modules                string
+	}
+
+	foo, err := yaml.Marshal(projectConfig.Modules)
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	t := tempProjectConfig{
+		Name:                   projectConfig.Name,
+		ShouldPushRepositories: projectConfig.ShouldPushRepositories,
+		Modules:                util.IndentString(string(foo), 2),
+	}
+
+	tmpl.Execute(os.Stdout, t)
+	result := tpl.String()
+	return result
 }
--- a/internal/config/projectconfig/init.go
+++ b/internal/config/projectconfig/init.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"fmt"
 	"io/ioutil"
-	"os"
 	"path"
 	"text/template"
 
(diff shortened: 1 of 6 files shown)


internal/config/projectconfig/init.go
```diff
@@ -59,3 +44,33 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
...
+		fmt.Println(err)
+	}
+
+	t := tempProjectConfig{
+		Name:                   projectConfig.Name,
+		ShouldPushRepositories: projectConfig.ShouldPushRepositories,
+		Modules:                util.IndentString(string(foo), 2),
+	}
+
+	tmpl.Execute(os.Stdout, t)
```
bmonkman, line 73: Needs to write to `tpl` instead of `os.Stdout`

internal/config/projectconfig/init.go
```diff
@@ -59,3 +44,33 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
 		exit.Fatal(fmt.Sprintf("Failed to create config file %s", constants.ZeroProjectYml))
 	}
 }
+
+func GetProjectFileContent(projectConfig ZeroProjectConfig) string {
+	var tpl bytes.Buffer
```
bmonkman, line 49: Give this a more meaningful name? (tpl vs tmpl is confusing)

internal/config/projectconfig/init.go
```diff
@@ -59,3 +44,33 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
...
+	tmpl := template.New("sa")
+	tmpl, err := tmpl.Parse(exampleConfig)
+	if err != nil {
+		exit.Fatal(fmt.Sprintf("Failed to parse the sample Zero module config file %s", constants.ZeroProjectYml))
+	}
+
+	type tempProjectConfig struct {
```
bmonkman, line 56: You can combine the definition and instantiation here by having an unnamed struct like this:
`t := struct {
   Name string 
    ... 
}{
    projectConfig.Name
    ...
}
`

internal/config/projectconfig/init.go
```diff
@@ -59,3 +44,33 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
 		exit.Fatal(fmt.Sprintf("Failed to create config file %s", constants.ZeroProjectYml))
 	}
 }
+
+func GetProjectFileContent(projectConfig ZeroProjectConfig) string {
```
bmonkman, line 48: This function needs to be integrated into init.

internal/util/util.go
```diff
@@ -113,3 +113,9 @@ func AppendProjectEnvToCmdEnv(envMap map[string]string, envList []string) []stri
 	}
 	return envList
 }
+
+// TODO: indent each line for the modules.
+func IndentString(content string, spaces int) string {
```
bmonkman, line 118: You could use a package like sprig for this which has easy to use support for templates. https://github.com/Masterminds/sprig

internal/config/projectconfig/init.go
```diff
@@ -59,3 +44,33 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
...
+		exit.Fatal(fmt.Sprintf("Failed to parse the sample Zero module config file %s", constants.ZeroProjectYml))
+	}
+
+	type tempProjectConfig struct {
+		Name                   string
+		ShouldPushRepositories bool
+		Modules                string
+	}
+
+	foo, err := yaml.Marshal(projectConfig.Modules)
```
bmonkman, line 62: Choose a better name than `foo` please.

internal/config/projectconfig/init.go
```diff
@@ -59,3 +44,33 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
...
+	if err != nil {
+		fmt.Println(err)
```
bmonkman, line 64: Should this fail here? Return this error?

internal/config/projectconfig/init.go
```diff
@@ -59,3 +44,33 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
 		exit.Fatal(fmt.Sprintf("Failed to create config file %s", constants.ZeroProjectYml))
 	}
 }
+
+func GetProjectFileContent(projectConfig ZeroProjectConfig) string {
+	var tpl bytes.Buffer
+	tmpl := template.New("sa")
+	tmpl, err := tmpl.Parse(exampleConfig)
+	if err != nil {
+		exit.Fatal(fmt.Sprintf("Failed to parse the sample Zero module config file %s", constants.ZeroProjectYml))
```
bmonkman, line 53: This is not a correct error message, you are parsing the config template.

internal/config/projectconfig/init.go
```diff
@@ -1,43 +1,28 @@
...
 
 	"github.com/commitdev/zero/internal/constants"
+	"github.com/commitdev/zero/internal/util"
 	"github.com/commitdev/zero/pkg/util/exit"
 	"gopkg.in/yaml.v2"
 )
 
-const exampleConfig = `name: %s
+// {{  .ShouldPushRepositories | printf "%q"}}
+const exampleConfig = `
```
bmonkman, line 18: Let's give this a better name. `zeroProjectConfigTemplate`?

internal/config/projectconfig/init.go
```diff
@@ -59,3 +44,33 @@ func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
 		exit.Fatal(fmt.Sprintf("Failed to create config file %s", constants.ZeroProjectYml))
 	}
 }
+
+func GetProjectFileContent(projectConfig ZeroProjectConfig) string {
+	var tpl bytes.Buffer
+	tmpl := template.New("sa")
```
bmonkman, line 50: You can chain these two lines. `tmpl, err := template.New("projectConfig").tmp.Parse(zeroProjectConfigTemplate)`

internal/config/projectconfig/init.go
dtoki: Yeah, It read better 👍🏽

internal/util/util.go
dtoki: I continued with a simple util function for indenting because I was not sure if sprig loops through each line to prefix each line with the right indentation.

bmonkman: It does: http://masterminds.github.io/sprig/strings.html

internal/config/projectconfig/init.go
dtoki: @bmonkman this is something I'm not 100% certain about and could use some clarity on, I've refactored it to exit with non-zero if serialization of the modules fails but I'm a bit unsure if I should rather propagate the error to the caller in this case `Init` too handle it there.

internal/util/util.go
```diff
@@ -114,8 +116,12 @@ func AppendProjectEnvToCmdEnv(envMap map[string]string, envList []string) []stri
 	return envList
 }
 
-// TODO: indent each line for the modules.
-func IndentString(content string, spaces int) string {
-	// TODO: implement me 
-	return "Testing indent string retrun"
+// IndentString will Add  x space char padding at the beginging of each line.
+func IndentString(content string, space int) string {
+	var result bytes.Buffer
```
bmonkman, line 121: You can do it without a byte buffer by just appending to a string.

`result += fmt.Sprintf(...`

internal/util/util.go
bmonkman: Fine to leave your function in there though, as it's already done.

internal/config/projectconfig/init.go
bmonkman: In most cases you should return the error to the caller to let it decide what to do.

bmonkman: Unless something absolutely fatal happened. In this case, seems better to return an error. Easier to test too.

internal/util/util.go
```diff
@@ -113,3 +114,13 @@ func AppendProjectEnvToCmdEnv(envMap map[string]string, envList []string) []stri
 	}
 	return envList
 }
+
+// IndentString will Add  x space char padding at the beginging of each line.
+func IndentString(content string, space int) string {
```
bmonkman, line 119: Mind pluralizing to "spaces"?

internal/config/projectconfig/init.go
```diff
@@ -47,15 +30,45 @@ func SetRootDir(dir string) {
 }
 
 func Init(dir string, projectName string, projectContext *ZeroProjectConfig) {
```
bmonkman, line 32: Might be worth refactoring this a bit.. Maybe rename this to `CreateProjectConfigFile`? 
And then since `GetProjectFileContent` probably isn't useful on its own, maybe merge the code into this function and then change the test to test this function and assert that the file was created and contains the proper content..
Right now you exposed `GetProjectFileContent` but it's not useful outside the package, and the test is testing internal functionality that was exposed only to be able to test it. 
An external test (in a `_test` file) should be testing only the exposed API of the package, and a package shouldn't expose more than it needs to.
Looks like you have a TODO for that below, but I would say combining these functions and then having a single test would be the way to go.

internal/config/projectconfig/init_test.go
```diff
@@ -7,26 +7,49 @@ import (
...
+		if !cmp.Equal(expectedConfig, resultConfig, cmpopts.EquateEmpty()) {
+			t.Errorf("projectconfig.ZeroProjectConfig.Unmarshal mismatch (-expected +result):\n%s", cmp.Diff(expectedConfig, resultConfig))
+		}
+	})
+
+	/* TODO: Test err is thrown when there is a failure?
+	t.Run("Should fail if modules are missing from project config", func(t *testing.T) {
+
+	})
+	*/
```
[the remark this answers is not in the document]
bmonkman: The better way to test would probably be for that function to return errors and move the fatal to the caller.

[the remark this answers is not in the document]
dtoki: 👍🏽, I initially thought about doing that but thought it might be an anti-pattern with cobra being the caller?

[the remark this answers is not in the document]
bmonkman: Well, our handler function is the caller, so that's fine.

PR #167, opened 2020-06-17
Add create command

bmonkman:


internal/context/init.go
```diff
@@ -106,8 +103,9 @@ func loadAllModules(moduleSources []string) map[string]moduleconfig.ModuleConfig
 			exit.Fatal("Unable to load module:  %v\n", err)
 		}
 		modules[mod.Name] = mod
+		mappedSources[mod.Name] = moduleSource
 	}
-	return modules
+	return modules, mappedSources
```
davidcheung, line 108: is it bad to just add source to the moduleConfig itself?

bmonkman: ModuleConfig represents the structure of the module file, so we don't want it in there.

internal/generate/generate_modules.go
```diff
@@ -9,67 +10,57 @@ import (
...
-	// }
-
-	// Prompt for module params and execute each of the generator modules
-	for _, mod := range templateModules {
-		// TODO: read zero-project.yml instead
+	// Make sure module sources are on disk
+	wg := sync.WaitGroup{}
+	wg.Add(len(projectConfig.Modules))
+	for _, mod := range projectConfig.Modules {
+		go module.FetchModule(mod.Files.Source, &wg)
```
davidcheung, line 30: Trying to understand how the go-getter stuff works
this will not download again if it's already on local?  for `Get()` on their [docs](https://godoc.org/github.com/hashicorp/go-getter#Get) they say
>"If dst already exists, Get will attempt to update it." 

and the `isLocal` function always receives the **remote-source** only, to do the `getter.Detect`, is getter.Get where the magic happens? 

https://github.com/commitdev/zero/blob/1a1ce930d6e87301004f09a58cb074164d24d245/internal/module/module.go#L23-L29

bmonkman: Yeah right now it may fetch it again, but so far it's fast enough that it hasn't been a problem. I want to go over this stuff again at some point anyway to have these temporary files in a global location rather than written to a local `tmp` directory. At that point we can put more thought into if/how we want to cache these files.

davidcheung: 👍  I see, was just trying to figure out what it does

bmonkman: Also the source location you provide can be local, not just remote, which is why we have that `isLocal` check there.

PR #174, opened 2020-06-22
Aws profile prompt

davidcheung:


internal/context/init.go
```diff
@@ -171,42 +171,87 @@ func getProjectPrompts(projectName string, modules map[string]moduleconfig.Modul
 	return handlers
```
bmonkman, line 171: Mind renaming context while you're in there? I'd like to get rid of that. I think this should probably be renamed to `internal/init` to match what we've been going for with other commands where the package is named after the command in `cmd`.

internal/context/prompts.go
```diff
@@ -10,17 +10,30 @@ import (
...
 	"github.com/commitdev/zero/pkg/util/exit"
 	"github.com/manifoldco/promptui"
 	"gopkg.in/yaml.v2"
 )
 
+// Constant to maintain prompt orders so users can have the same flow,
+// modules get downloaded asynchronously therefore its easier to just hardcode an order
+var AvailableVendorOrders = []string{"aws", "github", "circleci"}
+
+const AWS_PICK_PROFILE = "Existing AWS Profiles"
```
bmonkman, line 23: Go convention is to use camel case, and most likely you want to start with lower case since this should be unexported.

internal/context/prompts.go
```diff
@@ -31,6 +44,14 @@ func KeyMatchCondition(key string, value string) func(map[string]string) bool {
 	}
 }
 
+type CustomConditionSignature func(map[string]string) bool
+
+func CustomCondition(fn CustomConditionSignature) func(map[string]string) bool {
```
bmonkman, line 49: You should be able to specify the return type as `CustomConditionSignature` as well.

internal/context/prompts.go
```diff
@@ -150,24 +171,33 @@ func PromptModuleParams(moduleConfig moduleconfig.ModuleConfig, parameters map[s
...
 	// FIXME: what is a good way to dynamically modify partial data of a struct
 	// current just marashing to yaml, then unmarshaling into the base struct
 	yamlContent, _ := yaml.Marshal(promptsValues)
-	yaml.Unmarshal(yamlContent, &credentials)
-	return credentials
+	yaml.Unmarshal(yamlContent, &creds)
+
+	// Fill AWS credentials based on profile from ~/.aws/credentials
+	if val, ok := promptsValues["aws"]; ok {
+		if val["use_aws_profile"] == "Existing AWS Profiles" {
```
bmonkman, line 195: Use the const for comparison

pkg/credentials/credentials.go
```diff
@@ -37,6 +38,24 @@ func MakeAwsEnvars(cfg *projectconfig.ZeroProjectConfig, awsSecrets Secrets) []s
...
 
+func AwsCredsPath() string {
+	usr, err := user.Current()
+	if err != nil {
+		log.Fatal(err)
+	}
+	return filepath.Join(usr.HomeDir, ".aws/credentials")
+}
+
+func FillAWSProfileCredentials(credsPath string, profileName string, creds globalconfig.ProjectCredential) globalconfig.ProjectCredential {
```
bmonkman, line 49: Maybe just "Get" rather than "Fill"?

pkg/credentials/credentials.go
```diff
@@ -37,6 +38,24 @@ func MakeAwsEnvars(cfg *projectconfig.ZeroProjectConfig, awsSecrets Secrets) []s
...
+	awsCreds, err := credentials.NewSharedCredentials(credsPath, profileName).Get()
+	if err != nil {
+		log.Fatal(err)
+	}
+	creds.AWSResourceConfig.AccessKeyId = awsCreds.AccessKeyID
+	creds.AWSResourceConfig.SecretAccessKey = awsCreds.SecretAccessKey
+	return creds
+}
+
 func GetSecrets(baseDir string) Secrets {
```
bmonkman, line 59: Mind removing the old secrets.yml code?

pkg/credentials/credentials.go
bmonkman: Also, maybe split into two functions, one of which doesn't accept the path and just calls the other with the default of `AwsCredsPath()`? Seems like most likely the only time you would provide a different path would be for tests.

Issue #173, opened 2020-06-19
Add aws credentials to environment when executing commands for "execute" type parameters


bmonkman:
For example, the command that gets executed to get the current account id will use whatever account you executed zero as, not the account whose credentials you chose for the project.

PR #175, opened 2020-06-23
Integrate apply command

bmonkman:
Various changes to get the apply command working including passing more required env vars in and correcting some of the paths.


internal/config/globalconfig/global_config.go
```diff
@@ -50,6 +51,35 @@ func (p ProjectCredentials) Unmarshal(data []byte) error {
 	return nil
 }
 
+// AsEnvVars marshals ProjectCredential as a map of key/value strings suitable for environment variables
+func (p ProjectCredential) AsEnvVars() map[string]string {
```
bmonkman, line 55: @davidcheung This should help with #173

internal/apply/apply.go
```diff
@@ -2,54 +2,79 @@ package apply
...
+		fmt.Println(`Choose the environments to apply. This will create infrastructure, CI pipelines, etc.
+At this point, real things will be generated that may cost money!
+Only a single environment may be suitable for an initial test, but for a real system we suggest setting up both staging and production environments.`)
+		environments = promptEnvironments()
+	}
+
+	validateEnvironments(environments)
 
-	flog.Infof(":tada: Bootstrapping project %s. Please use the zero.[hcl, yaml] file to modify the project as needed. %s.", context.Name)
+	flog.Infof(":tada: Bootstrapping project %s. Please use the zero.yml file to modify the project as needed.", projectConfig.Name)
```
davidcheung, line 38: `zero-project.yml`

bmonkman:
Weird, dunno how this got closed. Must have been something I did in zenhub. Sorry.

tests/test_data/apply/zero-project.yml
```diff
@@ -0,0 +1,19 @@
+name: sample_project
+
+context:
```
dtoki, line 3: according to the new spec do we still support the `context:` field?

internal/apply/apply.go
```diff
@@ -2,54 +2,77 @@ package apply
...
 	"github.com/manifoldco/promptui"
 )
 
-// Apply will bootstrap the runtime environment for the project
-func Apply(dir string, applyConfigPath string, applyEnvironments []string) []string {
-	context := loadContext(dir, applyConfigPath, applyEnvironments)
+func Apply(rootDir string, configPath string, environments []string) {
+	if strings.Trim(configPath, " ") == "" {
+		exit.Fatal("config path cannot be empty!")
+	}
```
dtoki, line 25: Would it be better to return the error in this case and let the caller throw the error for consistency like in #130?

dtoki: Nvm, I just realized your comment was specific to if I wanted to test the error case so I guess that doesn't apply here since it's not a standard.

PR #180, opened 2020-06-24
Added dependsOn to module and project config, switched apply to apply…

bmonkman:
… as a graph traversal.


internal/config/projectconfig/project_config.go
```diff
@@ -63,9 +53,33 @@ func (c *ZeroProjectConfig) Print() {
 	pp.Println(c)
 }
 
-func NewModule(parameters Parameters, directory string, repository string, source string) Module {
+// GetDAG returns a graph of the module names used in this project config
+func (c *ZeroProjectConfig) GetDAG() dag.AcyclicGraph {
```
davidcheung, line 57: using AcyclicGraph is slick 🕶️  🕶️

bmonkman: Yeah it was nice to be able to handle dependencies in what is probably going to end up being the necessary way, but also be able to do it quite easily!

PR #181, opened 2020-06-24
vendor credentials is env-vars for execute

davidcheung:


internal/config/globalconfig/global_config.go
```diff
@@ -74,12 +76,36 @@ func gatherFieldTags(t reflect.Value, list map[string]string) map[string]string
...
+			if idx := strings.Index(opts, "omitempty"); idx != -1 && fieldValue.String() == "" {
+				continue
+			}
+			list[name] = fieldValue.String()
 		}
 	}
 	return list
 }
 
+func (p ProjectCredential) SelectedVendorsCredentialsAsEnv(vendors []string) map[string]string {
```
bmonkman, line 89: Mind adding this to Apply as well? 
Right now it's just loading all creds but I guess it would be nice to be more precise and only expose the necessary ones.
You'll need to load the module config to be able to get the required creds first though.

beiercai:
Hey @davidcheung , which command (init, apply etc..) is this ticket for?

davidcheung:
@beiercai this is for #173, just linked it in zenhub

PR #190, opened 2020-06-26
ZERO138 show summary at the end of Apply

cjhawkins:


internal/apply/apply.go
```diff
@@ -127,3 +128,39 @@ func validateEnvironments(applyEnvironments []string) {
 		}
 	}
 }
+
+func summarizeAll(dir string, projectConfig projectconfig.ZeroProjectConfig) {
+	flog.Infof("Your projects and infrastructure have been successfully created.  Here are some useful links and commands to get you started:")
```
cjhawkins, line 133: I'm not sure about the `Infof` statements in this function.. just wanted to get placeholders in to show text at the beginning and end of the summary.

internal/apply/apply.go
```diff
@@ -127,3 +128,39 @@ func validateEnvironments(applyEnvironments []string) {
...
+		}
+
+		modulePath := module.GetSourceDir(mod.Files.Source)
+		// Passed in `dir` will only be used to find the project path, not the module path,
+		// unless the module path is relative
+		if module.IsLocal(mod.Files.Source) && !filepath.IsAbs(modulePath) {
+			modulePath = filepath.Join(dir, modulePath)
+		}
+
+		envList = util.AppendProjectEnvToCmdEnv(mod.Parameters, envList)
```
bmonkman, line 159: Should also pass the environments in so the summary can show accurate info based on what was created.

cjhawkins:
The `apply` test will fail here until the module makefile changes are merged to master.  The test is trying to run `make summary` on each module.

PR #192, opened 2020-06-30
fix parsing project config missing value

davidcheung:
shouldPushRepositories flag was defaulting to false, so didn't realize the flag from yml file wasn't parsed 
and test case was expecting the default case, so didn't catch this


bmonkman:
Can you add more details?

PR #195, opened 2020-07-08
info: guide to obtain credentials

davidcheung:
adding some documentation on how to obtain credentials we prompt for (eg. github personal token / aws keys / circleci)


internal/init/prompts.go
```diff
@@ -29,6 +30,7 @@ type PromptHandler struct {
 	moduleconfig.Parameter
 	Condition CustomConditionSignature
 	Validate  func(string) error
+	Info      string
```
bmonkman, line 33: Maybe this should be part of `moduleconfig.Parameter` instead so that modules can provide more context for individual parameters?

davidcheung: right that make sense 👍 , didnt think of that

Issue #220, opened 2020-07-30
[bug] when project name has dashes, db-application-user isnt correctly created


davidcheung:
reason: the script made an assumption projectName and DB_NAME is the same, which isn't the case

because dashes get stripped out from the `DB_NAME`, the query ` --query "DBInstances[?DBName=='$PROJECT_NAME']` breaks when name includes dashes 
https://github.com/commitdev/zero-deployable-backend/blob/master/db-ops/create-db-user.sh#L6

problems:
- [ ] DB_ENDPOINT becomes null from the query in shell script
- [ ] DBNAME is also needed to connect to db in the kube deployment manifest
- [ ] .sql file also needs the DB_NAME i believe

Issue #216, opened 2020-07-23
validation for user prompts


davidcheung:
i think this is low effort high value
we already have validation feature, could be useful to run through the parameters we prompt and see if we can add some reasonable regex to avoid errors 

eg. some prompts such as `front/backend subdomain name`, we expect people to put `api.` which is not super clear and will lead to a failure down the road which will be frustrating

dtoki:
+1

PR #226, opened 2020-08-01
user prompt input validation

dtoki:
fix for #216 
related to #220

### DIFF
--- a/internal/init/prompts.go
+++ b/internal/init/prompts.go
@@ -88,8 +88,8 @@ func ValidateSAK(input string) error {
 	return nil
 }
 
-// ValidatePname validates Project Name field string input.
-func ValidatePname(input string) error {
+// ValidateProjectName validates Project Name field user input.
+func ValidateProjectName(input string) error {
 	// the first 62 char out of base64 and -
 	var pName = regexp.MustCompile(`^[A-Za-z0-9-]{1,15}$`)
 	if !pName.MatchString(input) {
@@ -200,6 +200,17 @@ func sanitizeParameterValue(str string) string {
 
 // PromptParams renders series of prompt UI based on the config
 func PromptModuleParams(moduleConfig moduleconfig.ModuleConfig, parameters map[string]string, projectCredentials globalconfig.ProjectCredential) (map[string]string, error) {
+	// map module field names to corresponding validate functions.
+	// not optimal solution as changing a field name in the EKS stack would render the map invalid.
+	m := map[string]func(string) error{
+		"productionHostRoot": validateRootDomain,
+		"stagingHostRoot":    validateRootDomain,
+
+		"productionFrontendSubdomain": validateSubDomain,
+		"productionBackendSubdomain":  validateSubDomain,
+		"stagingFrontendSubdomain":    validateSubDomain,
+		"stagingBackendSubdomain":     validateSubDomain,
+	}
 
 	credentialEnvs := projectCredentials.SelectedVendorsCredentialsAsEnv(moduleConfig.RequiredCredentials)
 	for _, promptConfig := range moduleConfig.Parameters {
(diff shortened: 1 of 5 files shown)


internal/init/prompts.go
```diff
@@ -88,6 +88,36 @@ func ValidateSAK(input string) error {
 	return nil
 }
 
+// ValidatePname validates Project Name field string input.
+func ValidatePname(input string) error {
+	// the first 62 char out of base64 and -
+	var pName = regexp.MustCompile(`^[A-Za-z0-9-]{1,15}$`)
```
dtoki, line 94: any thoughts on capping this at 15 chars. 

we briefly discussed possibly generating a shortened slug that `tf` can use instead of the project-name till that's done what do you think about capping this?

davidcheung: Does 15 chars work? no conflicts? 

If we were to limit the char length I would actually separate the return in 2 parts 
an if statement to check the len 
```go 
// perhaps something like this
if len(str) > maxLength {
  return errors.New("Exceeded maximum length blahblah")
}
if !pName.MatchString(input) { // other rules ...
```

internal/init/prompts.go
```diff
@@ -170,6 +200,17 @@ func sanitizeParameterValue(str string) string {
...
+	// not optimal solution as changing a field name in the EKS stack would render the map invalid.
+	m := map[string]func(string) error{
+		"productionHostRoot": validateRootDomain,
+		"stagingHostRoot":    validateRootDomain,
+
+		"productionFrontendSubdomain": validateSubDomain,
+		"productionBackendSubdomain":  validateSubDomain,
+		"stagingFrontendSubdomain":    validateSubDomain,
+		"stagingBackendSubdomain":     validateSubDomain,
+	}
```
dtoki, line 213: I'm going to move validation to the module side, allowing module variables to specify a regex to pick against or a predefined list of validation functions scales better. good suggestion @davidcheung 💡

internal/init/prompts.go
bmonkman: I do think 15 chars will be too limiting, but it should be fine for now.

internal/init/prompts.go
```diff
@@ -88,6 +88,36 @@ func ValidateSAK(input string) error {
...
+	// the first 62 char out of base64 and -
+	var pName = regexp.MustCompile(`^[A-Za-z0-9-]{1,15}$`)
+	if !pName.MatchString(input) {
+		return errors.New("Invalid project-name (cannot contain special chars except '-' & max len of 15)")
+	}
+	return nil
+}
+
+func validateRootDomain(input string) error {
+	// FIXME: this regex would still accept subdomains a.co.uk a permanent solution matching all TLDs
```
bmonkman, line 102: Let's just make it more permissive and allow TLDs with dots like your example of `a.co.uk`, but make sure that we have enough of a description during the prompt to give them an idea of why they need to enter a root domain and leave it up to them.

internal/init/prompts.go
dtoki: This is actually 4 chars too long see 👇

```shell
Error: expected length of name_prefix to be in the range (1 - 32), got dtoki-infra-015-production-us-west-2

  on .terraform/modules/production.eks.eks/terraform-aws-eks-12.1.0/cluster.tf line 112, in resource "aws_iam_role" "cluster":
 112: resource "aws_iam_role" "cluster" {
```

There is a 32 char cap on then aws I am role if we were to move forward, the max should be 11 chars (using the prod naming schema as the ceiling ) which really sucks, some options I can think of are:

1. change our infrastructure naming scheme for production  & staging.
 eg. -prod-us-west-2; -stage-us-west-2.
 this gives us 16 chars to work with.

2. create a unique shortened slug name that `tf` uses.

what do you folks suggest?

bmonkman: Yeah let's use prod and stage. Also even having the region in there is a bit overkill. Few startups will actually have multi-region setups which would require that to be part of the names.  Can you start a list in a google doc or somewhere else to track these places where we hit character limits in AWS?

dtoki: [aws limits doc](https://docs.google.com/document/d/1tNyWVx79H7N-R5JLq_7VXKceviQEFNqVoKf_hgHStHI/edit?usp=sharing)

internal/init/prompts.go
```diff
@@ -232,3 +259,22 @@ func appendToSet(set []string, toAppend []string) []string {
 	}
 	return set
 }
+
+// validation fuctions that field validation of type:fuction in zero-module.yaml can select
+// need to implement type:function flow in PromptModuleParams before use
+func validateRootDomain(input string) error {
```
bmonkman, line 265: Do we still need these now that the modules define their own regexes?

dtoki: right! I kept it as an example ref for when modules can select functions instead of using regexes, it's not necessary. I okay with removing it?

bmonkman: Yeah let's remove it.

PR #231, opened 2020-08-13
Zero root domain static site support

### LINKED ISSUE #229: Zero create step fails on templating a non-text file located in a directory
Running `zero create` crashes while templating non-text files; this happens if your zero template contains an image  / non-text file in a dir that isn't ignored.
 https://github.com/commitdev/zero/blob/b253a1ef75541b29c6aaec5edc93f31da9890256/internal/generate/generate_modules.go#L77-L88

**Possible Solutions:**
have another condition for 
- known binary file extensions
- read files header -> copy the file over if it's !type text.

related to #222

dtoki:
**Adds zero landing page registry**

- add a new repo to zero registry

**Don't template binary files.**

- Considering the definition of a binary file as "a computer file that is not a text file", they can be differentiated by searching for the text/plain MIME in its MIME hierarchy.

closes #229

### DIFF
--- a/internal/generate/generate_modules.go
+++ b/internal/generate/generate_modules.go
@@ -15,8 +15,9 @@ import (
 	"github.com/commitdev/zero/internal/module"
 	"github.com/commitdev/zero/internal/util"
 	"github.com/commitdev/zero/pkg/util/flog"
-
 	"github.com/commitdev/zero/pkg/util/fs"
+
+	"github.com/gabriel-vasile/mimetype"
 )
 
 // Generate accepts a projectconfig struct and renders the templates for all referenced modules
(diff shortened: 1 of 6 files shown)


internal/generate/generate_modules.go
```diff
@@ -65,9 +74,17 @@ type TemplateConfig struct {
 	isTemplate  bool
 }
 
+type BinFileConfig struct {
+	source      string
+	destination string
+	mineType    string
```
bmonkman, line 80: Typo

bmonkman: Any reason to have the mime type here? Or just because it was available?

dtoki: just because it's available; if we'd like to post-process on filter on it in the future.

bmonkman: Actually "isTemplate" in TemplateConfig is not even used anymore. How about we just have a struct with source and destination and use that for both return lists?

internal/generate/generate_modules.go
```diff
@@ -80,6 +97,29 @@ func newTemplates(moduleDir string, outputDir string, overwrite bool) []*Templat
...
+		if isBinary {
+			binFile = append(binFile, &BinFileConfig{
+				source:      path,
+				destination: fs.ReplacePath(path, moduleDir, outputDir),
+				mineType:    detectedMIME.String(),
+			})
+			continue
+		}
+
 		_, file := filepath.Split(path)
```
bmonkman, line 123: Mind removing this block? We don't use it anymore, as we just try to template everything in the repo (that's not binary)

PR #248, opened 2020-08-19
Add Wget dependency

### LINKED ISSUE #227: Wget should be a zero dependency
`zero check` should include wget as a dependency.

dtoki:
Adds Wget as a zero dependency

closes #227

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -124,7 +124,7 @@ var checkCmd = &cobra.Command{
 				command:    "terraform",
 				args:       []string{"version"},
 				regexStr:   `Terraform v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)`,
-				minVersion: "0.12.0",
+				minVersion: "0.13.0",
 				docsURL:    "https://www.terraform.io/downloads.html",
 			},
 			{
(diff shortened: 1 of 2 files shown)


cmd/check.go
```diff
@@ -143,6 +143,14 @@ var checkCmd = &cobra.Command{
 				minVersion: "2.17.1",
 				docsURL:    "https://git-scm.com/book/en/v2/Getting-Started-Installing-Git",
 			},
+			{
+				name:       "Wget\t\t",
+				command:    "wget",
+				args:       []string{"--version"},
+				regexStr:   `^GNU Wget (0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)`,
+				minVersion: "1.20.3",
```
davidcheung, line 151: looks like 1.20.3 is the latest one, people may not update their wget often
might be safer to go with some older versions as I doubt there's many breaking changes on wget

do you think it'll be good to allow some older versions too?

https://ftp.gnu.org/gnu/wget/?C=M;O=D

cmd/check.go
```diff
@@ -143,6 +143,14 @@ var checkCmd = &cobra.Command{
 				minVersion: "2.17.1",
 				docsURL:    "https://git-scm.com/book/en/v2/Getting-Started-Installing-Git",
 			},
+			{
```
davidcheung, line 146: Can you also bump the terraform to 0.13 please, the EKS stack's terraforms are now using some 0.13 features

bmonkman:
Can you also please add it as a dependency for brew in `.goreleaser.yml`?

cmd/check.go
dtoki: We're all probably using and testing on`1.20.3` but I'm comfortable supporting a minor version back `1.19.0` released in 2017

davidcheung: yeah 2012 definitely seems far back enough, I think as long as it covers most people's new computer cycle then its good 👍

PR #253, opened 2020-08-20
create: template to have access to files as well

davidcheung:
so templates can include directory/source/repository from https://github.com/commitdev/zero/blob/d53d483846f3faf8b77e75db64fed751ca0d4b83/internal/config/projectconfig/project_config.go#L34-L36


tests/test_data/generate/file_to_template.txt
```diff
@@ -1,2 +1,3 @@
 Name is {{.Name}}
 Params.test is {{.Params.test}}
+Files.test is {{.Files.Repository}}
```
bmonkman, line 3: should be "Files.Repository"

PR #278, opened 2020-09-15
Allow specifying local module paths in the init command

bmonkman:
To make it a little easier to work with local modules.
Eventually we will likely have a better solution for this as we dig more into the module ecosystem / registry idea, but for now this is a quick way to make it more flexible.


GrooveStomp:
Looks good to me!

GrooveStomp (approved):

PR #308, opened 2020-11-05
Adds new zero-module conditions parameter

GrooveStomp:
- Currently only supports 'ignoreFile' condition

### DIFF
--- a/internal/condition/condition.go
+++ b/internal/condition/condition.go
@@ -1,0 +1,35 @@
+package condition
+
+/*
+NOTE: To avoid cyclic dependencies, the actual struct datatype is not defined here.
+It is doubly defined - once in project_config.go and once in module_config.go.
+*/
+
+import (
+	"os"
+	"path"
+
+	"github.com/commitdev/zero/internal/config/projectconfig"
+)
+
+func Perform(cond projectconfig.Condition, mod projectconfig.Module) {
+	value, found := mod.Parameters[cond.MatchField]
+
+	// Exit if the condition isn't met.
+	if !found || value != cond.WhenValue {
+		return
+	}
+
+	// Okay, the condition was met, let's execute it.
+	switch cond.Action {
+	case "ignoreFile":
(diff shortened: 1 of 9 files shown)


internal/condition/condition.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	// Okay, the condition was met, let's execute it.
+	switch cond.Action {
+	case "ignoreFile":
+		ignoreFile(cond.Data, mod)
+	}
+}
+
+func ignoreFile(data []string, mod projectconfig.Module) {
```
davidcheung, line 30: maybe rename to `ignoreFiles`, or else it may be hard for module creators to understand `data` is a list of strings

davidcheung (approved):
nice! I can try using it for my user-auth changes in frontend once its merged

internal/condition/condition.go
GrooveStomp: That's fair.

internal/condition/condition.go
```diff
@@ -0,0 +1,35 @@
+package condition
+
+/*
```
bmonkman, line 3: Maybe add a comment here about what this is for? The naming is a big vague.

GrooveStomp: Literally none of the Go files have package or file-level comments at the top.
I say this only because I'm wondering if it should go _before_ the `package` statement, or after?

internal/condition/condition.go
bmonkman: We should be able to specify either files or directories in this list as well.

internal/condition/condition.go
bmonkman: It should be before.
https://blog.golang.org/godoc

GrooveStomp: Thanks. It's been a while since I was diligent about Go program documentation.
I went all-out on one project a couple years ago.

bmonkman (approved):
Nice, thanks!

PR #309, opened 2020-11-09
enable tests on main branch

davidcheung:
branch renamed and github actions had the wrong branch name

### DIFF
--- a/internal/condition/condition_test.go
+++ b/internal/condition/condition_test.go
@@ -1,6 +1,7 @@
 package condition_test
 
 import (
+	"encoding/base64"
 	"math/rand"
 	"os"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/condition/condition_test.go
```diff
@@ -12,7 +13,7 @@ import (
 func testSetup(paramKey, paramValue string) (string, projectconfig.Module) {
 	bytes := make([]byte, 15)
 	_, _ = rand.Read(bytes)
-	name := string(bytes[:])
+	name := string(base64.StdEncoding.EncodeToString(bytes[:]))
```
GrooveStomp, line 16: What was the issue here?

davidcheung: on Mac the test errors out with bytes as filename

GrooveStomp: :shrug:

GrooveStomp (approved):

PR #324, opened 2020-12-14
Validate zero modules

GrooveStomp:
Adds two private functions in 'internal/module_config':
- func (cfg ModuleConfig) collectMissing() []string
- func findMissing(reflect.Value, string, string, []string)

These are unexported functions that aid in introspecting a datastructure to see if it has all its expected data.
These functions are implicitly invoked via:
- func LoadModuleConfig(string) (ModuleConfig, error)

If there are errors then the program aborts with appropriate output.

### DIFF
--- a/internal/config/moduleconfig/module_config.go
+++ b/internal/config/moduleconfig/module_config.go
@@ -81,7 +81,7 @@ func LoadModuleConfig(filePath string) (ModuleConfig, error) {
 		flog.Errorf("%v is missing information", filePath)
 
 		for _, m := range missing {
-			flog.Errorf("\t %v\n", m)
+			flog.Errorf("\t %v", m)
 		}
 
 		log.Fatal("")
--- a/internal/config/moduleconfig/module_config.go
+++ b/internal/config/moduleconfig/module_config.go
@@ -66,9 +66,6 @@ func (cfg ModuleConfig) collectMissing() []string {
 func LoadModuleConfig(filePath string) (ModuleConfig, error) {
 	config := ModuleConfig{}
 
-	var required = []string{}
-	fmt.Printf("%+#v\n", required)
-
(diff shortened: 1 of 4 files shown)


internal/config/moduleconfig/module_config.go
```diff
@@ -48,16 +55,109 @@ type TemplateConfig struct {
...
+	findMissing(reflect.ValueOf(cfg), "", "", &missing)
+
+	return missing
+}
+
 func LoadModuleConfig(filePath string) (ModuleConfig, error) {
 	config := ModuleConfig{}
+
+	var required = []string{}
+	fmt.Printf("%+#v\n", required)
```
davidcheung, line 70: I assume this is not needed anymore? since we're treating non-omitempty as required

internal/config/moduleconfig/module_config.go
```diff
@@ -48,16 +55,109 @@ type TemplateConfig struct {
...
+// 2. YAML struct field metadata is sufficient to define whether an attribute is missing or not.
+//    That is, "yaml:foo,omitempty" tells us this is not a required field because we can omit it.
+// 3. Slices and arrays are assumed to be optional.
+//
+// As this function recurses through the datastructure, it builds up a string
+// path representing each node's path within the datastructure.
+// If the value of the current node is equal to the zero value for its datatype
+// and its struct field does *not* have a "omitempty" value, then we assume it
+// is missing and add it to the resultset.
+func findMissing(obj reflect.Value, path, metadata string, missing *[]string) {
```
davidcheung, line 108: 🕶️  👍

davidcheung (approved):

internal/config/moduleconfig/module_config.go
GrooveStomp: Yeah, for sure! Thanks.

PR #350, opened 2021-03-03
[SEMVER-MAJOR] condition support to module params

davidcheung:
feature changes includes:
- [x] removing global config [breaking change]
- [x] module config’s parameter to support conditions (all conditions must be true)
- module config’s parameter to support custom typed prompts (eg. AWS prompt - [x]ype)
- [x] module config parameter to support overwriting the env-var used when apply
- [x] module config parameter to support processing used parameters (`OmitFromProjectFile` default's to `false`)

behaviour changes includes:
- [x] prompts(function) to not return value, instead update map(can support multi-value) 
- [x] some parameter names improved (so the code is more readable)


bmonkman:
Go convention - can you please add comments on public functions starting with the name of the function?

internal/apply/apply.go
```diff
@@ -87,18 +87,24 @@ func applyAll(dir string, projectConfig projectconfig.ZeroProjectConfig, applyEn
...
+		envVarTranslationMap := modConfig.GetParamEnvVarTranslationMap()
+		envList = util.AppendProjectEnvToCmdEnv(mod.Parameters, envList, envVarTranslationMap)
 		flog.Debugf("Env injected: %#v", envList)
 		flog.Infof("Executing apply command for %s...", modConfig.Name)
 		util.ExecuteCommand(exec.Command("make"), modulePath, envList)
 		return nil
 	})
 }
 
+func getParameterDefinition(modConfig moduleconfig.ModuleConfig, field string) moduleconfig.Parameter {
```
bmonkman, line 99: Is this function not actually used anywhere?

bmonkman:
Can you add a new file in docs to capture the schema and usage of the module.yml now that it's getting more complex?

internal/apply/apply.go
davidcheung: Yeah looks like it, will remove

internal/config/moduleconfig/module_config.go
```diff
@@ -158,3 +178,38 @@ func findMissing(obj reflect.Value, path, metadata string, missing *[]string) {
...
+					moduleParams[parameterKey] = parameterValue
+				}
+			}
+		}
+	}
+	return moduleParams
+}
+
+// SummarizeConditions collects conditions from module-definition
+// then append the condition into the zero-project configuration
```
bmonkman, line 202: From the PoV of the user of this function it doesn't append into anything, it just returns a list of conditions.

internal/config/moduleconfig/module_config.go
```diff
@@ -158,3 +178,38 @@ func findMissing(obj reflect.Value, path, metadata string, missing *[]string) {
...
+func SummarizeConditions(module ModuleConfig) []projectconfig.Condition {
+	moduleConditions := []projectconfig.Condition{}
```
bmonkman, line 204: Not a big deal in this case and probably not worth a change, but in general if you know the final size of the slice ahead of time like this it's more efficient to pre-allocate it. For example, 
```
moduleConditions := make([]projectconfig.Condition{}, len(module.Conditions))
```
then as you loop through just assign each value like `moduleConditions[i] = newCond`

internal/config/projectconfig/project_config.go
```diff
@@ -29,6 +30,24 @@ type Module struct {
 	Conditions []Condition `yaml:"conditions,omitempty"`
 }
 
+// ReadVendorCredentialsFromModule uses parsed project-config's module
+// based on vendor parameter, retrieve the vendor's credential
+// for pre-defined functionalities (eg: Github api key for pushing repos to github)
+func ReadVendorCredentialsFromModule(m Module, vendor string) (error, string) {
+	// this mapping could be useful for module config as well
+	vendorToParamMap := map[string]string{
+		"github":   "githubAccessToken",
```
bmonkman, line 39: Where are these vendor strings going to come from?

pkg/credentials/credentials.go
```diff
@@ -8,30 +8,46 @@ import (
...
+	"github.com/commitdev/zero/internal/util"
 )
 
-func AwsCredsPath() string {
+type AWSResourceConfig struct {
+	AccessKeyID     string `yaml:"accessKeyId,omitempty" env:"AWS_ACCESS_KEY_ID,omitempty"`
+	SecretAccessKey string `yaml:"secretAccessKey,omitempty" env:"AWS_SECRET_ACCESS_KEY,omitempty"`
+}
+
+var GetAWSCredsPath = awsCredsPath
```
bmonkman, line 19: Why assign here rather than calling the function directly?

internal/util/util.go
```diff
@@ -142,3 +148,34 @@ func ItemInSlice(slice []string, target string) bool {
 	}
 	return false
 }
+
+// ReflectStructValueIntoMap receives a resource of struct type as
+// type AWSCreds struct{
+// 	AccessKeyID  string `yaml:"accessKeyId,omitempty"`
+// 	SecretAccessKey  string `yaml:"accessKeyId,omitempty"`
```
bmonkman, line 155: Assuming this should be `secretAccessKey` in the annotation.

internal/config/projectconfig/project_config.go
davidcheung: the function is actually only called for vcs (during `zero create`) right now, I didnt want to have the logic there on how to get the `github-api-key` from project config. 
so the `circleci` part is actually never used for now.

pkg/credentials/credentials.go
```diff
@@ -8,30 +8,46 @@ import (
 	"regexp"
 
 	"github.com/aws/aws-sdk-go/aws/credentials"
-	"github.com/commitdev/zero/internal/config/globalconfig"
+	"github.com/commitdev/zero/internal/util"
 )
 
-func AwsCredsPath() string {
+type AWSResourceConfig struct {
+	AccessKeyID     string `yaml:"accessKeyId,omitempty" env:"AWS_ACCESS_KEY_ID,omitempty"`
```
bmonkman, line 15: Why "yaml"? If this annotation is just to be used by the `ReflectStructValueIntoMap` function we might as well use our own identifier so it's clear that it's not supposed to be used in some other way to generate yaml.

davidcheung: No particular reason, mostly just because I reused the struct from before (global creds)
what should we change it to, should we just call it `zero`?

docs/project-definition.md
```diff
@@ -0,0 +1,27 @@
+### Project Definition: `zero-project.yml`
+Each project is defined by this project definition file, this manifest contains your project details, and is the source of truth for the templating(`zero create`) and provision(`zero apply`) steps. 
+
+`zero-project.yml` is created from `zero-init`, and defines which modules are part of the project and what each module's parameters are for the subsequence steps.
```
bmonkman, line 4:
```suggestion
Each project is defined by this file. This manifest reflects all the options a user chose during the `zero init` step. It defines which modules are part of the project, each of their parameters, and is the source of truth for the templating (`zero create`) and provision (`zero apply`) steps. 

_Note: This file contains credentials, so make sure it is not shared with others._
```

docs/module-definition.md
```diff
@@ -0,0 +1,55 @@
+## Module Definition: `zero-module.yml`
+Module definition defines the information needed for the module to run (`zero apply`).
+Also declares dependency used to determine the order of execution with other modules.
```
bmonkman, line 3:
```suggestion
This file is the definition of a Zero module. It contains a list of all the required parameters to be able to prompt a user for choices during `zero init`, information about how to template the contents of the module during `zero create`, and the information needed for the module to run (`zero apply`).
It also declares the module's  dependencies to determine the order of execution in relation to other modules.
```

docs/module-definition.md
```diff
@@ -0,0 +1,55 @@
...
+### Condition
+| Parameters   | Type         | Description                                                       |
+|--------------|--------------|-------------------------------------------------------------------|
+| `action`     | enum(string) | type of condition, currently supports [`KeyMatchCondition`]         |
+| `matchField` | string       | Allows you to condition prompt based on another parameter's value |
+| `WhenValue`  | string       | Matches for this value to satisfy the condition                   |
+| `data`       | list(string) | Supply extra data for condition to run                            |
+
+
+### Validation
```
bmonkman, line 49: Missing info on some of the other things like the `contitions` and `template` fields related to templating.

internal/init/prompts.go
```diff
@@ -126,9 +133,11 @@ func (p PromptHandler) GetParam(projectParams map[string]string) string {
 			exit.Fatal("Exiting prompt:  %v\n", err)
```
bmonkman, line 133: Mind adding a bit more info here? Maybe say that it's related to the module properties, and if possible, say which module the problem is with? I'm thinking for `CustomPromptHandler` for example, if the module specifies an invalid string, it's gonna be hard for a user to understand what's going on.

docs/module-definition.md
davidcheung: definitely good call to add this documentation, I looked at conditions and thought oh its there, but theres actually 2 types

bmonkman: Yeah the naming might be a bit confusing..

bmonkman (approved):

PR #351, opened 2021-03-12
prompt options to use map instead of list

davidcheung:
had to use mapSlice so the options can retain the order


bmonkman:
Nice! Can you please update the module definition documentation though?

bmonkman (approved):

PR #357, opened 2021-03-24
Version requirements

davidcheung:
allow modules to declare versions of Zero it would work with
this allows us to introduce breaking change and let users know the modules require newest features of Zero


Makefile
```diff
@@ -1,7 +1,7 @@
 VERSION = 0.0.1
 BUILD ?=$(shell git rev-parse --short HEAD)
 PKG ?=github.com/commitdev/zero
-BUILD_ARGS=-v -trimpath -ldflags=all="-X ${PKG}/cmd.appVersion=${VERSION} -X ${PKG}/cmd.appBuild=${BUILD}"
+BUILD_ARGS=-v -trimpath -ldflags=all="-X ${PKG}/version.AppVersion=${VERSION} -X ${PKG}/version.AppBuild=${BUILD}"
```
bmonkman, line 4: For real releases this actually gets set by .goreleaser.yml, not the Makefile.

docs/module-definition.md
```diff
@@ -2,14 +2,15 @@
...
-| `parameters`  | list(Parameter) | Parameters to prompt users                       |
+| Parameters    | type             | Description                                      |
+|---------------|------------------|--------------------------------------------------|
+| `name`        | string           | Name of module                                   |
+| `description` | string           | Description of the module                        |
+| `template`    | template         | default settings for templating out the module   |
+| `author`      | string           | Author of the module                             |
+| `icon`        | string           | Path to logo image                               |
+| `parameters`  | list(Parameter)  | Parameters to prompt users                       |
+| `zeroVersion` | string(go-semver)| Zero versions its compatible with                |
```
bmonkman, line 13: Maybe include a link to the go-version readme?

internal/config/moduleconfig/module_config.go
```diff
@@ -107,6 +138,10 @@ func LoadModuleConfig(filePath string) (ModuleConfig, error) {
 		log.Fatal("")
 	}
 
+	if !ValidateZeroVersion(config) {
+		constraint := config.ZeroVersion.Constraints.String()
+		return config, errors.New(fmt.Sprintf("Module's zero requirement not satisfied: expected %s  received: %s", constraint, version.AppVersion))
```
bmonkman, line 143: Let's add some more details, we are telling them a bit about what's wrong but not how to fix it, and it will be the user who is faced with this message.

internal/config/moduleconfig/module_config.go
```diff
@@ -214,3 +249,22 @@ func SummarizeConditions(module ModuleConfig) []projectconfig.Condition {
...
+
+// UnmarshalYAML Parses a version constraint string into go-version constraint during yaml parsing
+func (semVer *VersionConstraints) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var versionString string
+	err := unmarshal(&versionString)
+	if err != nil {
+		return err
+	}
+
+	constraints, constErr := goVerson.NewConstraint(versionString)
```
bmonkman, line 261: If the module.yml doesn't have a version specified we should probably just skip the check.

internal/config/moduleconfig/module_config.go
davidcheung: how about this 
image

bmonkman (approved):

PR #358, opened 2021-04-01
 apply: to check module requirements before run

davidcheung:
modules must now provide a `make check` in their root makefile
make check must exit 0 on all modules before proceeding to apply
this allows checks for module specific binaries or check tokens permissions


internal/apply/apply.go
```diff
@@ -33,6 +35,13 @@ Only a single environment may be suitable for an initial test, but for a real sy
 		environments = promptEnvironments()
 	}
 
+	flog.Infof(":tada: checking project %s. Please use the zero-project.yml file to modify the project as needed.", projectConfig.Name)
```
bmonkman, line 38: maybe `magnifying` instead of `tada` and remove the part about zero-project since it should appear later?

bmonkman:
Mind pasting text or posting a screenshot of what the final output looks like after this change?

internal/apply/apply.go
```diff
@@ -41,21 +50,26 @@ Only a single environment may be suitable for an initial test, but for a real sy
...
+	if err != nil {
+		return errors.New(fmt.Sprintf("Module summary failed: %s", err.Error()))
+	}
+	return err
 }
 
-func applyAll(dir string, projectConfig projectconfig.ZeroProjectConfig, applyEnvironments []string) {
-	environmentArg := fmt.Sprintf("ENVIRONMENT=%s", strings.Join(applyEnvironments, ","))
-
+func modulesWalkCmd(lifecycleName string, dir string, projectConfig *projectconfig.ZeroProjectConfig, cmdArgs []string, applyEnvironments []string) error {
```
bmonkman, line 68: Maybe `applyEnvironments` should be `environments` now that it's not apply specific??

davidcheung:
right now (before addressing any comments) it looks like 
image

bmonkman:
Hmm I wonder if we should be checking for return code 2 (a problem with make itself, like specifying a target that doesn't exist) and provide a slightly different message to indicate that it's likely a problem with the module itself.
Might be a little weird, as the function was made more generic, and this would only apply for the `check`

davidcheung:
```makefile
check:
	exit 1
```

Looks like even when i have the make target exits with error code 1,  the exec still exits with code 2, 
^ the snippet above still gives a code2 exit, so pretty much the exit code will always be 2? 

i tried to add a binary check as follow, the code returned from exec.cmd is still `2`
```makefile
REQUIRED_BINS := gh foobar

check:
	$(foreach bin, $(REQUIRED_BINS),\
		$(if $(shell command -v $(bin) 2> /dev/null),$(info Found `$(bin)`),$(error Please install `$(bin)`)))
```

but it can still show the error if the module is diligent with printing the error since we pipe it out 
image

bmonkman:
Hmm okay. In that case we could check stderr for the string "No rule to make target"

bmonkman (approved):

internal/apply/apply.go
```diff
@@ -41,21 +50,26 @@ Only a single environment may be suitable for an initial test, but for a real sy
 
 	flog.Infof("Infrastructure executor: %s", "Terraform")
 
-	applyAll(rootDir, *projectConfig, environments)
+	err = modulesWalkCmd("apply", rootDir, projectConfig, []string{"make"}, environments)
```
bmonkman, line 53: Will this stop on the first one that fails? We should really test all the modules, and then return the state / output of all the tests. It wouldn't be a great experience to have to run the app multiple times, installing stuff between runs.

davidcheung: yeah good point, right now it fails on the first error

internal/apply/apply.go
```diff
@@ -89,10 +110,58 @@ func applyAll(dir string, projectConfig projectconfig.ZeroProjectConfig, applyEn
...
+
+func getModuleOperationCommand(mod moduleconfig.ModuleConfig, operation string) (operationCommand []string) {
+	defaultCheck := []string{"make", "check"}
+	defaultApply := []string{"make"}
+	defaultSummary := []string{"make", "summary"}
+
+	switch operation {
+	case "check":
+		if mod.Commands.Check != "" {
+			operationCommand = strings.Split(mod.Commands.Check, " ")
```
bmonkman, line 145: You can use `strings.Fields()` for this. The only thing is it doesn't handle stuff like quotes, which we would need another library for, but it should be good enough. 
https://golang.org/pkg/strings/#Fields

internal/config/moduleconfig/module_config.go
```diff
@@ -22,14 +22,21 @@ type ModuleConfig struct {
 	Name                string
 	Description         string
 	Author              string
-	DependsOn           []string `yaml:"dependsOn,omitempty"`
+	Commands            ModuleCommands `yaml:"commands,omitempty"`
```
bmonkman, line 25: Can you add this to module-definition.md as well?

internal/apply/apply.go
davidcheung: is there an easy to to handle `|`s and `>` s  if users put some advance statement in the command?

bmonkman: Let's just go with Fields and encourage people to put any logic into a script

davidcheung:
so the new behavior looks like this, if the check uses a script it wont have the makefile noise in the output 

image

bmonkman:
Cool! Can we suppress the output as we are executing the command so we can just present Zero's nicer output?

internal/apply/apply_test.go
```diff
@@ -37,6 +30,13 @@ func TestApply(t *testing.T) {
 		content, err = ioutil.ReadFile(filepath.Join(tmpDir, "project2/project.out"))
 		assert.NoError(t, err)
 		assert.Equal(t, "baz: qux\n", string(content))
+
+	})
+
+	t.Run("Moudles runs command overides", func(t *testing.T) {
```
bmonkman, line 36: This and below say "Moudles"

bmonkman: That does have a nice ring to it though..

internal/util/util.go
```diff
@@ -68,19 +73,39 @@ func ExecuteCommand(cmd *exec.Cmd, pathPrefix string, envars []string) {
...
-		log.Fatalf("Starting command failed: %v\n", err)
+		return err
 	}
 
 	go func() {
 		_, errStdout = io.Copy(os.Stdout, stdoutPipe)
 	}()
 	go func() {
-		_, errStderr = io.Copy(os.Stderr, stderrPipe)
+		strErrStreams := []io.Writer{errContent}
```
bmonkman, line 83: is this supposed to be `stdErrStreams`?

davidcheung: sorry yeah, should have paid more attention with all these typos

bmonkman (approved):

PR #391, opened 2021-07-09
feat: Add an info box showing more context during the info command

bmonkman:


internal/init/prompts_test.go
```diff
@@ -17,7 +17,7 @@ func TestGetParam(t *testing.T) {
 	t.Run("Should execute params without prompt", func(t *testing.T) {
 		param := moduleconfig.Parameter{
 			Field:   "account-id",
-			Execute: "echo \"my-acconut-id\"",
+			Execute: "echo \"my-account-id\"",
```
davidcheung, line 20: 😅 👍

davidcheung (approved):
LGTM

internal/init/prompts_test.go
bmonkman: This wasn't the fix, just noticed it while I was in there. 😂

PR #401, opened 2021-08-06
doc-site: core-concept and life-cycle explanation

davidcheung:
- moving concepts above reference  ( i think it makes more sense to be read first)?
- broken links on overview page
- added 2 pages to explain how the steps work (removed module life cycle as its very coupled to project life-cycle)


doc-site/docs/concepts/core-concepts.md
```diff
@@ -3,3 +3,15 @@ title: Core Concepts
...
+
+The Project manifest(`zero-project.yml`) is the source of truth for the commands (`create` and `apply`). It determines where to fetch the modules, the execution order of modules, whether it will push your project to version control, and other project information. You can provision both staging and production environment using the same manifest to ensure the environments are reproducible and controlled.
+
+## Module
+A module is useful bundle of code and/or resources that can be templated out with the Project Manifest, then executes a provision flow, this could be templating out terraform infrastructure as code then provisioning the resources, or creating a backend application and deploying it.
+
+A module is defined by it's **Module manifest**(`zero-module.yml`) in it's root folder, it contains all the parameters it requires, and declares it's requirements and execution.
+
+Modules can declare it's dependencies, for example a backend that will be deployed can declare its dependency on the infrastructure repository, so that it will execute the infrastructure module's flow before the backend itself.
```
bmonkman, line 17:
```suggestion
A project defines a set of **modules** to use, along with a full set of config parameters for each module which were entered by the user during the `zero init` step. These config options are stored in `zero-project.yml`.

The project manifest (`zero-project.yml`) is the source of truth for the commands `create` and `apply`. It determines from where to fetch the modules, the execution order of modules, whether it will push your project to version control, module configuration, and other project information. Both staging and production environments are provisioned using the same manifest to ensure the environments are reproducible and controlled.

## Module
A module is a useful bundle of code and/or resources that can be templated out during the `create` step, then executes a provisioning flow. This could be templating out terraform infrastructure as code then provisioning the resources, or creating a backend application, making API calls to set up a build pipeline, and then deploying it.

A module is defined by the **module manifest** (`zero-module.yml`) in its root folder. It contains all the parameters required to render the templated files (during `zero create`) and execute any provisioning steps, and declares it's requirements and the commands to execute during provisioning (`zero apply`).

Modules can declare their dependencies, for example a backend that will be deployed can declare its dependency on the infrastructure repository so that the infrastructure will already exist by the time we want to deploy to it.
```

doc-site/docs/concepts/project-life-cycle.md
```diff
@@ -3,3 +3,27 @@ title: Project Life Cycle
...
+
+## zero apply
+`zero apply` setups up the final steps of your provisioning, it by default runs the `make` command and modules can specify to change this behavior. Zero expects a two step process of which it will run `check` then `make` to setup the module.
+
+### Check
+`check` is ran for all the modules in your project before attempting to do the `run` step, so if one dependent's `check` fails it will not start the actual provisioning for any of the modules, this is useful to check for missing binaries or API token permissions.
+
+### Apply
+The run step is by default `make` and can be overriden in the module definition. Run should be the one time setup procedures that allow the module to function.
+For example in the infrastructure repository this would be **running terraform**, and for the backend repository this could be **setting up the webhooks for your circleCI to connect to your github**.
```
bmonkman, line 29:
```suggestion
## zero init
The goal of the `init` step is to create the project manifest (`zero-project.yml`).

`zero init` will fetch each **module** defined in `zero-project.yml` from their remote repository, and prompt the user through a series of questions to fill in parameters required by each module. In this phase, the module definition will be parsed and provide defaults, options, and extra context to guide users through filling in their project details.

:::note
It's recommended to review the `zero-project.yml` and make adjustments as needed before running `zero create` and `zero apply`.
:::

## zero create
`zero create` is run in the same folder as `zero-project.yml`. It will template out each module specified in the project manifest as the basis of your repositories, then push them to your version control repository (Github).

During the `create` step, Zero will also conditionally include or exclude certain sets of files, as defined by each module. For example, it will not scaffold the authentication examples if you opted not to use this feature.

## zero apply
`zero apply` is the provisioning step that starts to make real-world changes. By default, it runs a command defined by the module to check for any dependencies, and then runs a command to actually apply any changes.

### Check
`check` is run for all the modules in your project before attempting to do the `run` step, so if a dependent's `check` fails it will not start the actual provisioning for any of the modules. This is useful to check for missing binaries or API token permissions before starting to apply any changes.

### Apply
By default, the run step is to execute `make` in the root of the module, but that can be overridden in the module definition. Run should be the one-time setup commands that allow the module to function.
For example, in the infrastructure repository, this would be to **run terraform**, and for the backend repository, this could be to  **make API calls to CircleCI to set up your build pipeline**.
```

bmonkman (approved):

PR #403, opened 2021-08-09
remove custom stack

davidcheung:
custom stack is not useful for anything atm, as theres no docs
on how to put together your own stack and just causes confusion


bmonkman (approved):
Thanks!

PR #410, opened 2021-08-30
fix: missing projectName context in prompt:execute

davidcheung:
reference in modules for `projectName` doesn't work anymore in places like https://github.com/commitdev/zero-backend-node/blob/8db8db64bce4d7772ea6de2d235aebe97edc5c2a/zero-module.yml#L127-L128

Looks like used to be appended from the [`projectCredentials`](https://github.com/commitdev/zero/commit/7f4c7a34a21426df0b09e1018bce4f05ed468150#diff-e1c5b6a139d58abff4f87ec5e4598a25b582dc09de7a2743649868f3bc9e5b34L124)


bmonkman (approved):
Nice catch, thanks!

PR #420, opened 2021-10-06
feat: Added support for modules being able to supply a default value for a parameter when the condition is not met. Also improved documentation.

bmonkman:


bmonkman:
Removed the docs dir which is now deprecated.

PR #435, opened 2021-11-19
Update issue #434: trim \n out

### LINKED ISSUE #434: zero create: defaultBranch cmd output from git includes linebreak
looks like the command comes back with a `\n` in the content, will have to trim this out or creation can fail

```
DEBUG[2021-11-05T14:43:21-04:00] Running (git) command in ./frontend0818-test, []string{"push", "-u", "origin", "master\n"}
```

aserputov:
Resolve #434.

### DIFF
--- a/internal/vcs/create-git-repos.go
+++ b/internal/vcs/create-git-repos.go
@@ -17,14 +17,14 @@ func InitializeRepository(repositoryUrl string, githubApiKey string) {
 	var err error
 	ownerName, repositoryName, err := parseRepositoryUrl(repositoryUrl)
 	if err != nil {
-		fmt.Printf("error creating repository: %s\n", err.Error())
+		fmt.Printf("error creating repository: %s", err.Error())
 		return
 	}
 	flog.Debugf("Initialized repo: %s/%s", ownerName, repositoryName)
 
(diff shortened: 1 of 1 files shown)


davidcheung:
Hi @aserputov 
Sorry didn't really make the issue clear originally, the problem is the `output` of the `getInitDefaultBranch()` from exec comes back with a `\n` so when its appended to the list of cmds  it also have the linebreak in there 
```go
args:        []string{"push", "-u", "origin", initDefaultBranch},
```
 
https://github.com/commitdev/zero/blob/4937a10f52856db4703f5aca646223680cf484f1/internal/vcs/create-git-repos.go#L160-L162

aserputov:
@davidcheung I hope I'm correct. After checking how the file is working, I think the "%s" might create a new line.  
https://github.com/commitdev/zero/blob/4937a10f52856db4703f5aca646223680cf484f1/internal/vcs/create-git-repos.go#L199 - L201

davidcheung:
I suspect it is from the `git config --get init.defaultBranch` command, for printing nicely in stdout they may have appended a linebreak, we could probably trim it out from the `output`

PR #443, opened 2021-12-21
Fix: Use correct github actions page url

vnbhattarai:
Noticed github actions url was pointing to https://google.ca. This fixes it :)


davidcheung:
Good catch, thanks!!

davidcheung (approved):

PR #448, opened 2022-01-21
Show error if template is not good

pmoranga:
Print error if template is not good.


davidcheung (approved):
Thanks! definitely much better than panic and nil pointer trace 🎉 
image

Issue #469, opened 2022-06-09
ability to override or append to registry


davidcheung:
Problem: 
Currently registry is hardcoded, the struct is very simple right now 
https://github.com/commitdev/zero/blob/587d6f575532f5cdc1c5e6796ac4e8aab5ce78f5/internal/registry/registry.go#L3-L7
that means to build a module u must also build your own zero binary. 

Potential solution: 
Maybe we can allow something like ~/.zero-registry.yaml, we can just add some annotations to the struct and  parse the config, allowing users to customize registry for their own use

JeanPinzon:
Working on it

JeanPinzon:
Hey everyone! I created the PR #470 trying to solve it. Please let me know if I should change anything. Thanks!

bmonkman:
PR merged and v0.2.5 released: https://github.com/commitdev/zero/releases/tag/v0.2.5
Thanks again!

PR #470, opened 2022-06-13
Ability to override registry

JeanPinzon:
Issue: #469 

- Add `registry.yaml` with the default stacks;
- Add the parameter `registry-file-path` to override the file path on the command `init`. It can be a local or a github path.


JeanPinzon:
#469

JeanPinzon:
> Thanks for your contribution!! I wonder going forward if we still need to use the `local-module-path` given we can override the entire registry, but we can remove it in a different PR if we find ourselves not using it anymore

I'm not sure if `local-module-path` is just used on `registry`, but if so, it would be interesting to remove it and put the entire URL on the registry file. With this approach, it would be possible to have stacks with templates from different GitHub accounts.

davidcheung:
Yeah i think the registry file "should" cover all the use cases, go-getter allows a pretty wide range of paths
it is a breaking change though (which only affects development on modules) 

@bmonkman what do u think about removing the local-module-path entirely?

bmonkman:
Let's just leave the local module path for now and remove it separately later if we truly end up not using it. I think it's still probably useful as-is when using default values.

bmonkman (approved):
Looks great, thanks @JeanPinzon!

